PR #5, opened 2015-11-05
Issue/CCP-104

nischalsamji:
Updated Code to fix Issue CCP-104

### DIFF
--- a/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/AddUserController.java
+++ b/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/AddUserController.java
@@ -4,6 +4,8 @@ import javax.servlet.http.HttpServletRequest;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Controller;
+import org.springframework.ui.ModelMap;
+import org.springframework.web.bind.annotation.ModelAttribute;
 import org.springframework.web.bind.annotation.RequestMapping;
 
 import edu.asu.conceptpower.users.IUserManager;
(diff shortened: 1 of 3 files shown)


Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/AddUserController.java
```diff
@@ -26,26 +28,41 @@
...
 
-		User user = new User(req.getParameter("username"),
-				req.getParameter("password"));
-		user.setName(req.getParameter("fullname"));
-		user.setEmail(req.getParameter("email"));
-		usersManager.addUser(user);
+		User newUser = new User(user.getUser(), user.getPw());
+		newUser.setName(user.getName());
+		newUser.setEmail(user.getEmail());
+		if(usersManager.findUser(newUser.getUser())== null)
```
jdamerow: We're following the same coding style as in class: http://diging.github.io/course-book/2015/07/10/code-style.html
Make sure you're using curly brackets and format them correctly please.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/AddUserController.java
```diff
@@ -26,26 +28,41 @@
...
+			usersManager.addUser(user);
+		else
```
jdamerow: You probably should do this kind of validation in a validator rather than the controller.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/AddUserController.java
```diff
@@ -26,26 +28,41 @@
...
 	 * 
+	 * @param model
+	 * 				Holds generic information about the model
+	 *            
+	 * 
 	 * @return String to redirect user to add user page
 	 */
 	@RequestMapping(value = "auth/user/add")
-	public String addUser() {
+	public String addUser(ModelMap model) {
```
jdamerow: I know you didn't name this method, but let's call it something else. Maybe "preparePage" or something?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/AddUserController.java
nischalsamji: Can I use the existing validation package and make a new User validator class?

jdamerow:
see comments

PR #9, opened 2015-11-13
Story/ccp 117

mkarthik90:
[CCP-117] Fixed Review comments

### DIFF
--- a/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptEditController.java
+++ b/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptEditController.java
@@ -209,6 +209,7 @@ public class ConceptEditController {
 		ConceptEntry concept = conceptManager.getConceptEntry(conceptid);
 		List<ConceptEntry> synonyms = new ArrayList<ConceptEntry>();
 		String synonymIds = concept.getSynonymIds();
+		
 		if (synonymIds != null) {
 			String[] ids = synonymIds.trim().split(Constants.SYNONYM_SEPARATOR);
 			if (ids != null) {
--- a/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptEditController.java
+++ b/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptEditController.java
@@ -10,19 +10,19 @@ import java.util.Map;
 import javax.servlet.http.HttpServletRequest;
 
 import org.codehaus.jettison.json.JSONException;
-import org.json.simple.JSONArray;
-import org.json.simple.JSONObject;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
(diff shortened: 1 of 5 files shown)


Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptEditController.java
```diff
@@ -216,13 +206,79 @@ public String confirmlEdit(@PathVariable("id") String id,
...
+	private void buildJSONWithSynonymObject(ConceptEntry synonym, StringBuffer jsonStringBuilder) {
+		jsonStringBuilder.append("{");
+		jsonStringBuilder.append("\"Id\":\"" + synonym.getId() + "\"");
+		jsonStringBuilder.append(",");
+		jsonStringBuilder.append("\"Word\":\"" + synonym.getWord() + "\"");
+		jsonStringBuilder.append(",");
+		String description = synonym.getDescription().replaceAll("\"", "'");
+		jsonStringBuilder.append("\"Description\":\"" + description + "\"");
+		jsonStringBuilder.append(",");
+		jsonStringBuilder.append("\"SynonymObject\":\"" + synonym + "\"");
```
jdamerow: Do you actually need this? It doesn't seem to be used. And if you don't need it, why don't you just use the above method (buildJSONWithoutSynonymObject) and remove this one? Also, why are some labels capitalized ("Word", "Id", in this method), but not others ("id", "word" in previous method)?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptEditController.java
```diff
@@ -216,13 +206,79 @@ public String confirmlEdit(@PathVariable("id") String id,
 				}
 			}
 		}
-
 		ConceptEntry[] arraySynonyms = new ConceptEntry[synonyms.size()];
+
 		int i = 0;
+		StringBuffer jsonStringBuilder = new StringBuffer("{");
```
jdamerow: Ideally you want all of the JSON code out of this method. Any reason you have this here and not in it's own method?

jdamerow:
see comments

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptEditController.java
mkarthik90: Yes thats true. I had it as a separate method because in future if someone wants to add few attributes to JSON and also these synonym attributes it would be easier to call this method after adding the required attributes. 
If you want me to add all these code snippet into one private method I can do that.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptEditController.java
mkarthik90: I have made both the attributes same. But the difference between these two methods is that For each iteration I have added a synonymObject, where as in previous method I dont have a synonym Object. And also previous method contains POS and this method does not contain pos attribute

PR #11, opened 2015-11-19
Story/ccp 117

mkarthik90:
[CCP-117] Fixed the code review comments

### DIFF
--- a/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/bean/ConceptEditBean.java
+++ b/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/bean/ConceptEditBean.java
@@ -7,11 +7,12 @@ import java.util.Map;
 import org.springframework.stereotype.Component;
 
 import edu.asu.conceptpower.core.ConceptType;
+import edu.asu.conceptpower.core.POS;
 
 @Component
-public class ConceptEditBean {
+public class ConceptEditBean implements POS {
 
-	private Map<String, String> possMap = new LinkedHashMap<String, String>();
+	private Map<String, String> posMap = new LinkedHashMap<String, String>();
 	private String word;
 	private String concept;
 	private String selectedPosValue;
@@ -94,15 +95,6 @@ public class ConceptEditBean {
 		this.selectedPosValue = selectedPosValue;
 	}
 
-	public Map<String, String> getPossMap() {
-		possMap.put("noun", "Nouns");
-		possMap.put("verb", "Verbs");
-		possMap.put("adverb", "Adverb");
-		possMap.put("adjective", "Adjective");
-		possMap.put("other", "Other");
-		return possMap;
-	}
-
 	public String getSelectedListName() {
 		return selectedListName;
 	}
@@ -110,7 +102,6 @@ public class ConceptEditBean {
 	public void setSelectedListName(String selectedListName) {
 		this.selectedListName = selectedListName;
 	}
-
 
 	public String getSelectedTypeId() {
 		return selectedTypeId;
@@ -152,10 +143,6 @@ public class ConceptEditBean {
 		this.conceptList = conceptList;
 	}
 
-	public void setPossMap(Map<String, String> possMap) {
-		this.possMap = possMap;
-	}
-
 	public ConceptType[] getTypes() {
 		return types;
 	}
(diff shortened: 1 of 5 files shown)


Conceptpower+Spring/src/main/java/edu/asu/conceptpower/bean/ConceptEditBean.java
```diff
@@ -0,0 +1,182 @@
...
+	public List getConceptEntryList() {
+		return conceptEntryList;
+	}
+
+	public void setConceptEntryList(List conceptEntryList) {
+		this.conceptEntryList = conceptEntryList;
+	}
+
+	public Map<String, String> getPosMap() {
+		posMap.put(edu.asu.conceptpower.core.POS.NOUN, POS.NOUN);
```
jdamerow: why do you use the full path here? is it a different POS than the other one?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/bean/ConceptEditBean.java
```diff
@@ -0,0 +1,182 @@
...
+		posMap.put(edu.asu.conceptpower.core.POS.ADVERB, POS.ADVERB);
+		posMap.put(edu.asu.conceptpower.core.POS.ADJECTIVE, POS.ADJECTIVE);
+		return posMap;
+	}
+
+	public void setPosMap(Map<String, String> posMap) {
+		this.posMap = posMap;
+	}
+
+}
```
jdamerow: formatting

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptEditController.java
```diff
@@ -216,13 +184,71 @@ public String confirmlEdit(@PathVariable("id") String id,
...
+			}
+
+			jsonStringBuilder.append("{");
+			jsonStringBuilder.append("\"Id\":\"" + syn.getId() + "\"");
+			jsonStringBuilder.append(",");
+			jsonStringBuilder.append("\"Word\":\"" + syn.getWord() + "\"");
+			jsonStringBuilder.append(",");
+			jsonStringBuilder.append("\"Description\":\"" + syn.getDescription().replaceAll("\"", "'") + "\"");
+
+			if (posRequired == true) {
```
jdamerow: there is not really a reason to sometimes not have the POS; even if you don't show it (which is unlikely) you can add it here; would be easier and cleaner.

Conceptpower+Spring/src/main/webapp/WEB-INF/views/auth/concepts/ConceptEditView.jsp
```diff
@@ -4,248 +4,248 @@
...
-						$("#addedSynonyms").html(html);
+                        var htmlValue = '<tr id='+eachSynonym.Id+'><td align="justify"><font size="2">'
+                                + '<a onclick="synonymTemporaryRemove(\''
+                                + eachSynonym.Id
+                                + '\')">Remove</a>'
+                                + '</font></td>';
+                        htmlValue += '<td align="justify"><font size="2">'
+                                + eachSynonym.Word + '</font></td>';
+                        htmlValue += '<td align="justify"><font size="2">'
+                                + eachSynonym.Description + '</font></td></tr>';
```
jdamerow: You might want to think about refacting this to use jquery's append method.

Conceptpower+Spring/src/main/webapp/WEB-INF/views/auth/concepts/ConceptEditView.jsp
```diff
@@ -4,248 +4,248 @@
...
+                                                    + '</font></td></tr>';
+                                            // arr.push(eachSynonym.SynonymObject);
 
-									}
-								});
-					});
-</script>
+                                        }
+                                        // document.getElementById("conceptEntryList").value = arr;
+                                        html += '</tbody></table>';
```
jdamerow: same here

jdamerow:
see comments

Conceptpower+Spring/src/main/webapp/WEB-INF/views/auth/concepts/ConceptEditView.jsp
mkarthik90: @jdamerow  Should I refactor htmlvalue? I am not sure which variable I need to refactor.

jdamerow: looking at this again, I think I confused it with bootstrapTable append. You already do what I had in mind. We might want to think to switch to Bootstrap for the layout, and then we can make this much prettier. but for now, let's leave it with this.

PR #15, opened 2015-11-19
Story/ccp 92

jdamerow:


jdamerow:
CCP-110 does not seem to be solved; I still get an error saying that something went horribly wrong.

image

jdamerow:
the error appears after I try to change the concept list of an entry; there are no dupliate concept lists;

PR #16, opened 2015-11-19
Included the methods for handling the details link.

mkarthik90:
[CCP-89] Added the details function to fetch the details for wordnet concepts


jdamerow:
ok to test

jdamerow:
The wordnet Id is missing from the details view:
image
 
There should be one.

PR #20, opened 2015-11-24
Story/ccp 117

mkarthik90:
[CCP-117] Review Comments have been fixed. Need to switch to Bootstrap for the layout. Will be handled in separate story.

### DIFF
--- a/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptEditController.java
+++ b/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptEditController.java
@@ -257,14 +257,14 @@ public class ConceptEditController {
 	
 	private void buildJSONWithoutSynonymObject(StringBuffer jsonStringBuilder, ConceptEntry syn) {
 		jsonStringBuilder.append("{");
-		jsonStringBuilder.append("\"id\":\"" + syn.getId() + "\"");
+		jsonStringBuilder.append("\"Id\":\"" + syn.getId() + "\"");
 		jsonStringBuilder.append(",");
-		jsonStringBuilder.append("\"word\":\"" + syn.getWord() + "\"");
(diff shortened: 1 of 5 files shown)


Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptEditController.java
```diff
@@ -216,13 +184,64 @@ public String confirmlEdit(@PathVariable("id") String id,
...
+	 * 
+	 * @param synonymid
+	 *            Holds the ID of a synonym
+	 * @param model
+	 *            A generic model holder for Servlet
+	 * @return synonym details
+	 * @throws JSONException
+	 */
+	@RequestMapping(method = RequestMethod.GET, value = "getConceptAddSynonyms")
+	public ResponseEntity<String> getSynonymRows(@RequestParam("synonymid") String synonymid, ModelMap model)
```
jdamerow: do you need the "throws JsonException" here? I don't see it being thrown.

Conceptpower+Spring/src/main/webapp/WEB-INF/views/auth/concepts/ConceptEditView.jsp
```diff
@@ -4,248 +4,248 @@
...
+                                .ajax({
+                                    type : "GET",
+                                    url : "${pageContext.servletContext.contextPath}/getConceptEditSynonyms",
+                                    data : {
+                                        conceptid : conceptid
+                                    },
+                                    success : function(response) {
+                                        var border = response.length > 0 ? 1
+                                                : 0;
+                                        var html = '<table border="'+ border +'" width="400" id="addedSynonymsTable"><thead></thead><tbody>';
```
jdamerow: I think I would put the table into the base html, and then just add/append the rows.

jdamerow:
please resolve the conflicts.

PR #23, opened 2015-11-24
Story/ccp 105

mkarthik90:
[CCP-105] New request


jdamerow:
ok to test

jdamerow:
nope, this able to do that. The first meaning of horse wasn't shown in the table naymore, but the second one was, I could add it several times.
image

PR #27, opened 2015-11-25
Story/ccp 118

mkarthik90:
[CCP-118] Included model attributes and validator

### DIFF
--- a/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptSearchController.java
+++ b/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptSearchController.java
@@ -7,11 +7,18 @@ import javax.servlet.http.HttpServletRequest;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Controller;
 import org.springframework.ui.ModelMap;
+import org.springframework.util.CollectionUtils;
+import org.springframework.validation.BindingResult;
+import org.springframework.validation.annotation.Validated;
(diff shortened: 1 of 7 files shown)


Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptSearchController.java
```diff
@@ -24,43 +31,47 @@
...
+            BindingResult results) {
+
+        if (results.hasErrors()) {
+            return "conceptsearch";
+        }
+
+        ConceptEntry[] found = conceptManager.getConceptListEntriesForWord(conceptSearchBean.getWord(),
+                conceptSearchBean.getPos().toLowerCase().trim());
+        List<ConceptEntryWrapper> foundConcepts = wrapperCreator.createWrappers(found);
+        conceptSearchBean.setFoundConcepts(foundConcepts);
```
jdamerow: merge line 68 and 69

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptSearchController.java
```diff
@@ -24,43 +31,47 @@
...
+        if (CollectionUtils.isEmpty(foundConcepts)) {
+            results.rejectValue("foundConcepts", "no.searchResults");
+        }
```
jdamerow: I wouldn't do it that way. It's not a validation error that there are no results. Just show a message "Nothing matching your search query." where the table usually would be whenever the results are 0.

jdamerow:
see comments

PR #30, opened 2015-12-01
Story/ccp 113

mkarthik90:
[CCP-113] Resolved Bean Definition


Conceptpower+Spring/src/main/java/edu/asu/conceptpower/validation/ConceptTypeAddValidator.java
```diff
@@ -0,0 +1,49 @@
...
+    @Override
+    public void validate(Object obj, Errors errors) {
+        ValidationUtils.rejectIfEmptyOrWhitespace(errors, "typeName", "required.type_name");
+        ValidationUtils.rejectIfEmptyOrWhitespace(errors, "typeDescription", "required.type_description");
+
+        if (!errors.hasErrors()) {
+            ConceptTypeAddForm conceptTypeAddForm = (ConceptTypeAddForm) obj;
+            // If the name is same as old name then user has not changed the
+            // concept type name
+            if (!conceptTypeAddForm.getOldTypeName().equalsIgnoreCase(conceptTypeAddForm.getTypeName())) {
```
jdamerow: why do you do this check here?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptTypeEditController.java
```diff
@@ -27,84 +33,98 @@
...
+    public String prepareEditType(@PathVariable("typeid") String typeid, ModelMap model,
+            @ModelAttribute("conceptTypeAddForm") ConceptTypeAddForm conceptTypeAddForm) {
+
+        ConceptType type = typeManager.getType(typeid);
+
+        conceptTypeAddForm.setTypeName(type.getTypeName());
+        conceptTypeAddForm.setOldTypeName(type.getTypeName());
+        conceptTypeAddForm.setTypeDescription(type.getDescription());
+        conceptTypeAddForm.setMatches(type.getMatches());
+        conceptTypeAddForm.setSelectedType(type.getSupertypeId().trim());
```
jdamerow: What's "selected type"? Shouldn't that be "supertype"?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/validation/ConceptTypeAddValidator.java
mkarthik90: @jdamerow I think we discussed about this in the previous pull request comments. This condition was included to make sure duplicate type names are not being added to the system while editing the concept type name. 

Say user first tries to change the concept type name from Test to Test1, the first condition will pass through since 
conceptTypeAddForm.getOldTypeName() which is equal to Test  and 
conceptTypeAddForm.getTypeName() which is equal to Test1.

In the next condition the client.findType(conceptTypeAddForm.getTypeName()) != null , is true and record will be returned for type name Test1 and he will thrown an error on the screen unique type name is required.


Now user changes his mind and does not want to change the Type name and changes the type name to Test (to the old type name that was existing for this type). So in this scenario the first condition will not pass through
!conceptTypeAddForm.getOldTypeName().equalsIgnoreCase(conceptTypeAddForm.getTypeName()).

If the first condition is not available, everytime when the user updates the system by having the same type name, say Test in our scenario he will be thrown an error message "Type name should be unique". This scenario occurs only in editing the concept type names.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptTypeEditController.java
mkarthik90: @jdamerow  Selected type is nothing but super type in jsp. Pls find the below snippet.

<td>Super Type</td>
			<td><form:select path="selectedType">
					<form:option value="" label="Select one" />
					<c:forEach items="${conceptTypeAddForm.types}" var="typesDropDown">
						<form:option value="${typesDropDown.value}">${typesDropDown.value}</form:option>
					</c:forEach>
				</form:select></td>

mkarthik90: @jdamerow If you are fine with the comments you can merge the changes. I did not make any code changes for this branch now

jdamerow: My question here is, why it is "selectedType"? It should be supertype as it is the same, right? Makes it very confusing if you change the variable name from supertype to selected type, or is there a reason it is called "selected type" and not "supertype"?

jdamerow:
ok to test

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptTypeEditController.java
mkarthik90: @jdamerow  Yes I have changed from selectedType to superType and committed the changes.

PR #32, opened 2015-12-01
Story/ccp 117

mkarthik90:
[CCP-117] Validated the removal of synonym. Fixed Jquery issue. Verified store modified concepts button


Conceptpower+Spring/src/main/java/edu/asu/conceptpower/bean/ConceptEditBean.java
```diff
@@ -0,0 +1,182 @@
...
+import java.util.List;
+import java.util.Map;
+
+import org.springframework.stereotype.Component;
+
+import edu.asu.conceptpower.core.ConceptType;
+import edu.asu.conceptpower.core.POS;
+
+@Component
+public class ConceptEditBean implements POS {
```
jdamerow: why are you implementing this interface?

mkarthik90: Sorry implements is not required here so removed it. Just importing the package will do

PR #33, opened 2015-12-01
Story/ccp 121

mkarthik90:
[CCP-121] Introduced buildnumber and tested. Included pullrequest with an empty check


jdamerow:
ok to test

PR #43, opened 2015-12-08
Changes for type uri

mkarthik90:
[CCP-74] Tested the flow using this URL and was able to fetch the results successfully. http://localhost:8080/conceptpower/rest/ConceptSearch?type_uri=http://www.digitalhps.org/types/TYPE_6e11d4cd-3be1-4d6f-8eeb-c8673d68fd90&operator=and

### DIFF
--- a/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/rest/ConceptSearch.java
+++ b/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/rest/ConceptSearch.java
@@ -19,6 +19,7 @@ import edu.asu.conceptpower.core.ConceptEntry;
 import edu.asu.conceptpower.core.ConceptType;
 import edu.asu.conceptpower.core.IConceptManager;
 import edu.asu.conceptpower.db4o.TypeDatabaseClient;
+import edu.asu.conceptpower.util.URIHelper;
 import edu.asu.conceptpower.xml.XMLConceptMessage;
 import edu.asu.conceptpower.xml.XMLMessageFactory;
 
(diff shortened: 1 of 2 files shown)


Conceptpower+Spring/src/main/java/edu/asu/conceptpower/rest/ConceptSearch.java
```diff
@@ -56,15 +60,19 @@
...
+            if (key.trim().equals(SearchParamters.OPERATOR) && !queryParams.get(key)[0].trim().isEmpty())
+                operator = queryParams.get(key)[0].trim();
+            else {
+                // If the value is typeURI trim and add it to typeId
+                if (key.trim().equalsIgnoreCase(SearchFieldNames.TYPE_URI)) {
+                    searchFields.put("type_id", uriHelper.getTypeId(queryParams.get(key)[0]));
+                } else {
+                    searchFields.put(key.trim(), queryParams.get(key)[0]);
+                }
+            }
```
jdamerow: I'd turn this into and if/elseif/else statement instead of a nested if/else.

also, I think the indentations are wrong?

jdamerow:
ok to test

jdamerow:
@mkarthik90 would you take a look at this please? I don't think that you broke it but Nischal's last branch that i forgot to test before merging :(

jdamerow:
http://diging.asu.edu:8091/job/Conceptpower_deploy_on_push_to_develop/29/console

jdamerow:
retest this please

PR #45, opened 2015-12-08
Story/ccp 78

mkarthik90:
[CCP-78] Changes for adding delete functionality in concept search home screen

### DIFF
--- a/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptDeleteController.java
+++ b/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptDeleteController.java
@@ -111,7 +111,7 @@ public class ConceptDeleteController {
 		return "/auth/conceptlist/concepts";
 	}
 	
-	@RequestMapping(value = "auth/conceptlist/deleteconcepts/{id}", method = RequestMethod.GET)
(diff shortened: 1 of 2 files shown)


Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptDeleteController.java
```diff
@@ -110,5 +110,13 @@ public String confirmlDelete(@PathVariable("id") String id, ModelMap model) {
 		model.addAttribute("result", foundConcepts);
 		return "/auth/conceptlist/concepts";
 	}
-
+	
+    @RequestMapping(value = "auth/conceptlist/deleteconcepts/{id}", method = RequestMethod.GET)
+    public String deleteConcept(@PathVariable("id") String id, ModelMap model) {
+        ConceptEntry concept = conceptManager.getConceptEntry(id);
+        concept.setDeleted(true);
+        conceptManager.storeModifiedConcept(concept);
```
jdamerow: This logic should be done in the conceptmanager. Your controller should just call delete on the manager, and the manager knows what it has to do (not deleting the concept, but setting delete to true).

jdamerow:
see comment

PR #47, opened 2015-12-08
Adjusted width of td.

mkarthik90:
[CCP-71] Adjusted width of td in concept list screen


jdamerow:
ok to test

jdamerow:
please do the same for the user table.

PR #48, opened 2015-12-09
Story/ccp 78

mkarthik90:
[CCP-78] Made review comments changes and included delete manager

### DIFF
--- a/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/core/impl/DeleteConceptManager.java
+++ b/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/core/impl/DeleteConceptManager.java
@@ -1,0 +1,16 @@
+package edu.asu.conceptpower.core.impl;
+
+import org.springframework.stereotype.Service;
+
+import edu.asu.conceptpower.core.ConceptEntry;
+import edu.asu.conceptpower.core.IDeleteConceptManager;
+
+@Service
+public class DeleteConceptManager  implements IDeleteConceptManager{
+
+    @Override
+    public void setDelete(ConceptEntry entry) {
+        entry.setDeleted(true);
+    }
+
+}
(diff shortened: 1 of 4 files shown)


Conceptpower+Spring/src/main/java/edu/asu/conceptpower/core/impl/DeleteConceptManager.java
```diff
@@ -0,0 +1,16 @@
...
+import org.springframework.stereotype.Service;
+
+import edu.asu.conceptpower.core.ConceptEntry;
+import edu.asu.conceptpower.core.IDeleteConceptManager;
+
+@Service
+public class DeleteConceptManager  implements IDeleteConceptManager{
+
+    @Override
+    public void setDelete(ConceptEntry entry) {
```
jdamerow: this method should be part of the conceptmanger. dont' create a sepearte class, just add it to the exsiting one.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptDeleteController.java
```diff
@@ -110,5 +114,13 @@ public String confirmlDelete(@PathVariable("id") String id, ModelMap model) {
 		model.addAttribute("result", foundConcepts);
 		return "/auth/conceptlist/concepts";
 	}
-
+	
+    @RequestMapping(value = "auth/conceptlist/deleteconcepts/{id}", method = RequestMethod.GET)
+    public String deleteConcept(@PathVariable("id") String id, ModelMap model) {
+        ConceptEntry concept = conceptManager.getConceptEntry(id);
+        deleteConceptManager.setDelete(concept);
+        conceptManager.storeModifiedConcept(concept);
```
jdamerow: you would simply call
{code}
conceptManager.deleteConcept(id);
{code}

the conceptManager would retrieve the concept, set deleted to true and then store the concept.

jdamerow:
see comments

PR #51, opened 2015-12-10
Ccp 81

mkarthik90:
[CCP-81] Made changes to create wordnet wrapper and concept search.

### DIFF
--- a/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/core/ConceptEntry.java
+++ b/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/core/ConceptEntry.java
@@ -63,6 +63,8 @@ public class ConceptEntry implements Serializable {
 	
 	
 	private String broadens;
+	
+	private String uri;
 
 	@SearchField(fieldName = SearchFieldNames.CREATOR)
 	private String creatorId;
@@ -337,6 +339,14 @@ public class ConceptEntry implements Serializable {
 			return false;
 		return true;
 	}
+
+    public String getUri() {
+        return uri;
+    }
+
+    public void setUri(String uri) {
+        this.uri = uri;
+    }
 
 	
 }
--- a/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/core/ConceptEntry.java
+++ b/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/core/ConceptEntry.java
@@ -64,8 +64,6 @@ public class ConceptEntry implements Serializable {
 	
 	private String broadens;
 	
-	private String uri;
-
 	@SearchField(fieldName = SearchFieldNames.CREATOR)
 	private String creatorId;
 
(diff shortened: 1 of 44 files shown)


jdamerow:
please resolve conflicts

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/wrapper/ConceptEntryWrapper.java
```diff
@@ -17,6 +19,8 @@
  * @author Julia Damerow
  * 
  */
+@Component("conceptEntryWrapper")
+@Scope(value = "prototype")
 public class ConceptEntryWrapper implements Serializable {
```
jdamerow: Didn't we discuss this? That should be just a simple class, no scope, no component. And then the uri is a property that is set in the wrapper creator.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/core/ConceptEntry.java
```diff
@@ -330,6 +332,14 @@ public boolean equals(Object obj) {
...
 	}
 
+    public String getUri() {
+        return uri;
+    }
+
+    public void setUri(String uri) {
+        this.uri = uri;
+    }
+
```
jdamerow: I don't think you need this property here. Where do you use it?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/wrapper/ConceptEntryWrapper.java
mkarthik90: @jdamerow  Yes Julia, without having conceptentrywrapper as a spring bean we cannot autowire the bean in ConceptWrapperAddController search method(). If that is fine I will remove the @Component and create the object through new keyword.

jdamerow:
what about the other comment?

mkarthik90:
@jdamerow

jdamerow:
can we put that in the wrapper? we don't want to store the URI in the database where it would end up if it's part of the conceptentry.

mkarthik90:
@jdamerow  I have made the changes for URI and verified

jdamerow:
please resolve conflicts

mkarthik90:
@jdamerow  Resolved merge conflicts. Could you please verify and merge branch

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/wrapper/ConceptEntryWrapper.java
jdamerow: I was so sure I responded to that. Hm, well... apparently not.

Yes, remove component and create a new object with new.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/wrapper/impl/ConceptEntryWrapperCreator.java
```diff
@@ -31,7 +35,13 @@
 
 	@Autowired
 	private IUserManager usersManager;
+	
+	@Autowired
+    private IURIHelper helper;
+	
+	private BeanFactory beanFactory;
```
jdamerow: I don't think you still need that, do you?

PR #53, opened 2015-12-10
Changes for displaying pullrequest values

mkarthik90:
[CCP-122] Removed the default value from pom and made changes in skeleton to make the condition work as expected.


jdamerow:
ok to test

PR #55, opened 2015-12-10
Made code changes for CCP-90

nischalsamji:
Included Check box to make a user as admin while adding the user


jdamerow:
ok to test

PR #58, opened 2015-12-14
Story/ccp 117

mkarthik90:
[CCP-117] Issue resolved on authentication admin user. Moved the details to property file.

### DIFF
--- a/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/users/impl/UsersManager.java
+++ b/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/users/impl/UsersManager.java
@@ -1,8 +1,10 @@
 package edu.asu.conceptpower.users.impl;
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
@@ -10,6 +12,10 @@ import java.util.UUID;
 import javax.annotation.PostConstruct;
 
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.PropertySource;
+import org.springframework.core.env.AbstractEnvironment;
+import org.springframework.core.env.Environment;
+import org.springframework.core.io.support.ResourcePropertySource;
 import org.springframework.security.crypto.bcrypt.BCrypt;
 import org.springframework.stereotype.Service;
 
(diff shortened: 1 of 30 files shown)


Conceptpower+Spring/src/main/resources/user.properties
```diff
@@ -0,0 +1 @@
+admin=admin1234,ROLE_CP_ADMIN,enabled
```
jdamerow: there needs to be at least one user coming from the pom.xml, which means the two maven variables (admin.user and admin.pw) need to be added.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/users/impl/UsersManager.java
```diff
@@ -30,10 +34,21 @@
...
+	
+	
+	
 	@PostConstruct
-	public void init() {
+	public void init() throws IOException {
 		admins = new HashMap<String, String>();
+		
+		Properties prop = new Properties();
+		InputStream is = this.getClass().getResourceAsStream("/user.properties");
```
jdamerow: use spring to read the properties file

jdamerow: use spring to read the properties file; you can annotate classes to read specific files (qstore might have an example, or quadriga too I believe).

jdamerow:
see comments

mkarthik90:
@jdamerow  Resolved merge conflicts and included spring tags

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/users/impl/UsersManager.java
```diff
@@ -24,17 +30,41 @@
...
+	
+	
+    @PostConstruct
+    public void init() throws IOException {
+        admins = new HashMap<String, Object>();
+        for (Iterator it = ((AbstractEnvironment) env).getPropertySources().iterator(); it.hasNext();) {
+            Object source = (Object) it.next();
+            if (source instanceof ResourcePropertySource) {
+                ResourcePropertySource propertySource = (ResourcePropertySource) source;
+                admins.putAll(((ResourcePropertySource) propertySource).getSource());
```
jdamerow: not sure if that's the best way of doing it. Why don't you call getPropertyNames() on the propertysource and iterate over the returned array of username here and add them to the admins map right away (loose the loop in line 61).

Conceptpower+Spring/src/main/resources/user.properties
jdamerow: this is still missing

jdamerow:
see comments

PR #67, opened 2015-12-25
Ccp 126

mkarthik90:
[CCP-126] Fixed review comments.


Conceptpower+Spring/src/test/java/edu/asu/conceptpower/core/ConceptManagerTest.java
```diff
@@ -15,146 +13,146 @@
...
+        ConceptEntry newConcept = new ConceptEntry();
+        newConcept.setConceptList("list1");
+        newConcept.setCreatorId("testuser");
+        newConcept.setDescription("description");
+        newConcept.setPos("noun");
+        newConcept.setWord("test new");
+        newConcept.setWordnetId("WID-1");
+
+        String id = managerToTest.addConceptListEntry(newConcept);
+        Assert.assertNotNull(id);
```
jdamerow: you probably also want to test that the store method of the dbmanager is callsed

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/core/ConceptManagerTest.java
```diff
@@ -15,146 +13,146 @@
...
+        newConcept.setWord("test new");
+        newConcept.setWordnetId("WID-1");
+
+        String id = null;
+        try {
+            id = managerToTest.addConceptListEntry(newConcept);
+        } catch (DictionaryDoesNotExistException e) {
+            // test passes
+        }
+        Assert.assertNull(id);
```
jdamerow: you don't need to test that the id is null as this is not part of the logic of the manager class. just test for the exception to be thrown. you should be able to add that to the Test annotation 
```
@Test(expected=Exception_class_here.class)
```

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/core/ConceptManagerTest.java
```diff
@@ -15,146 +13,146 @@
...
+        newConcept.setWordnetId("WID-1");
+
+        String id = null;
+        try {
+            id = managerToTest.addConceptListEntry(newConcept);
+        } catch (DictionaryModifyException e) {
+            // test passes
+        }
+
+        Assert.assertNull(id);
```
jdamerow: same here

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/core/impl/ConceptListManagerNullTest.java
```diff
@@ -0,0 +1,25 @@
...
+import org.junit.Before;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
+
+import edu.asu.conceptpower.core.ConceptList;
+import edu.asu.conceptpower.db4o.IConceptDBManager;
+
+public class ConceptListManagerNullTest {
```
jdamerow: what is this class doing?

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/core/impl/ConceptListManagerTest.java
```diff
@@ -0,0 +1,78 @@
...
+        assertNotNull(list);
+        assertEquals(1, list.size());
+    }
+
+    @Test
+    public void getConceptListTest() {
+        String listName = "First List";
+        ConceptList conceptList = conceptListManager.getConceptList(listName);
+        assertNotNull(conceptList);
+        assertEquals(listName, conceptList.getConceptListName());
```
jdamerow: instead of this, I would test for equality of the returned concept list and the concept list you defined in the init method

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/core/impl/ConceptListManagerTest.java
```diff
@@ -0,0 +1,78 @@
...
+        String listName = "Second List";
+        ConceptList conceptList = conceptListManager.getConceptList(listName);
+        assertNull(conceptList);
+    }
+
+    @Test
+    public void checkExistingListForTrue() {
+        String listName = "First List";
+        boolean isExists = conceptListManager.checkExistingConceptList(listName);
+        assertEquals(true, isExists);
```
jdamerow: instead you can simply use 
```
assertTrue(...)
```
I believe

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/core/impl/ConceptListManagerTest.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Test
+    public void checkExistingListForFalse() {
+        String listName = "Second List";
+        boolean isExists = conceptListManager.checkExistingConceptList(listName);
+        assertEquals(false, isExists);
```
jdamerow: same here, assertFalse

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/core/impl/ConceptTypesManagerTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+public class ConceptTypesManagerTest {
+
+    @Mock
+    private TypeDatabaseClient client = Mockito.mock(TypeDatabaseClient.class);
+
+    @InjectMocks
+    private ConceptTypesManager conceptTypesManager;
+
+    ConceptType type = new ConceptType();
```
jdamerow: should be private

jdamerow:
see comments; just a few things

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/core/ConceptManagerTest.java
mkarthik90: @jdamerow  Should i write a separate test class for dbmanager and test the methods of dbmanager. Because the dbmanager is just storing the values to the db and not returning anything. Or do you want me to mock dbmanager (store method)

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/core/ConceptManagerTest.java
```diff
@@ -113,13 +114,16 @@ public void testAddConceptListEntry() throws DictionaryDoesNotExistException, Di
...
 		newConcept.setWordnetId("WID-1");
-		
+
 		String id = managerToTest.addConceptListEntry(newConcept);
 		Assert.assertNotNull(id);
 	}
-	
+
+	@Rule
+	public ExpectedException expectedEx = ExpectedException.none();
```
jdamerow: I think I'd move that declaration to the beginning of the class with all the other fields.

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/core/ConceptManagerTest.java
```diff
@@ -127,19 +131,17 @@ public void testAddConceptListEntryWrongDict() throws DictionaryModifyException
...
-		} 
-		
+		expectedEx.expect(DictionaryDoesNotExistException.class);
+		id = managerToTest.addConceptListEntry(newConcept);
 		Assert.assertNull(id);
 	}
-	
+
+	@Rule
+	public ExpectedException expected = ExpectedException.none();
```
jdamerow: do you really need this? can't you use the one you already declared?

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/core/ConceptManagerTest.java
jdamerow: You already mock the dbmanager, right? You should be able to verify with Mockito that the correct method was called: https://mockito.googlecode.com/hg-history/1.5/javadoc/org/mockito/Mockito.html

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/core/ConceptManagerTest.java
```diff
@@ -147,14 +149,8 @@ public void testAddConceptListEntryToWordnet() throws DictionaryDoesNotExistExce
...
-		try {
-			id = managerToTest.addConceptListEntry(newConcept);
-		} catch (DictionaryModifyException e) {
-			// test passes
-		}
-		
-		Assert.assertNull(id);		
+		expectedEx.expect(DictionaryModifyException.class);
+		String id = managerToTest.addConceptListEntry(newConcept);
+		Assert.assertNull(id);
```
jdamerow: You're still asserting that the id is null, which it shouldn't because an exception is thrown so there is no return value, right?

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/core/impl/ConceptListManagerTest.java
```diff
@@ -0,0 +1,77 @@
...
+    public void getAllConceptListsTest() {
+        List list = conceptListManager.getAllConceptLists();
+        assertNotNull(list);
+        assertEquals(1, list.size());
+    }
+
+    @Test
+    public void getConceptListTest() {
+        String listName = "First List";
+        ConceptList concept = conceptListManager.getConceptList(listName);
```
jdamerow: bad variable name

jdamerow:
just a few more minor things

jdamerow:
retest this please

PR #79, opened 2016-02-26
Story/new layout

mkarthik90:
Fixed review comments


jdamerow:
ok to test

jdamerow:
What about the concept edit page?
image

jdamerow:
Delete concept page is missing new layout for buttons.
image

jdamerow:
Can we have the menu entries in the drop down menus shaded as in the regular bootstrap drop-down:
image
http://getbootstrap.com/components/#dropdowns

jdamerow:
Let's use the following icons for the menu entries:
Conceptlist
- Concept List -> `<i class="fa fa-list"></i>` => This should also say "Show Concept Lists"
- Add New Concept List -> `<i class="fa fa-plus-circle"></i>`
add a separator here
- Add New Concept -> `<i class="fa fa-plus-circle"></i>`
- Add New Wordnet Concept Wrapper -> `<i class="fa fa-plus-circle"></i>`

Concept Types:
- Show Concept Types -> `<i class="fa fa-tags"></i>`
- Add New Type -> `<i class="fa fa-plus-circle"></i>`

Users:
- Manage Users -> `<i class="fa fa-users"></i>`
- Add New user -> `<i class="fa fa-plus-circle"></i>`

jdamerow:
On Concept Lists, Concept Types, and Users page, use the full width of the page if there is no side menu.

jdamerow:
On concept list page, the html is not rendered:
image

jdamerow:
Add new concept type page needs new layout.

jdamerow:
On user management page, put a new user button above the table (same on types page and concept list page).

jdamerow:
Add new user link doesn't work. It just reloads the page I am on.

jdamerow:
On user management page, replace edit password icon with this one `<i class="fa fa-key"></i>`.

jdamerow:
Edit user and edit password page need new layout.

jdamerow:
Delete user page needs new layout. Also, the column with the field name is way too long. Should only be as long as the labels. And name row has the colon before the name instead of after the label. There is also a space between username label and colon.
image

jdamerow:
delete type page needs new layout

PR #80, opened 2016-03-05
Ccp 128

mkarthik90:
[CCP-128] [story-newlayout] Review comments fixes


Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/lucene/impl/LuceneDAO.java
```diff
@@ -0,0 +1,97 @@
...
+
+    ObjectContainer luceneClient;
+
+    private String dbPath;
+
+    @PostConstruct
+    public void init() {
+
+        EmbeddedConfiguration config = Db4oEmbedded.newConfiguration();
+        config.common().objectClass(Timestamp.class).translate(new TSerializable());
```
jdamerow: For what do you  need this?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/web/LuceneIndexController.java
```diff
@@ -0,0 +1,97 @@
...
+        try {
+            manager.deleteIndexes();
+            manager.indexConcepts();
+            bean = dao.getTotalNumberOfWordsIndexed();
+        } catch (LuceneException | IllegalArgumentException | IllegalAccessException ex) {
+            bean.setMessage(ex.getMessage());
+            return bean;
+        }
+        bean.setMessage("Indexed successfully");
+        System.out.println("Outside");
```
jdamerow: remove sysouts

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/web/LuceneIndexController.java
```diff
@@ -0,0 +1,97 @@
...
+     * This method deletes all the index in Lucene and recreates all the index
+     * from Wordnet and from CCP db40 database
+     * 
+     * @param req
+     * @param principal
+     * @param model
+     * @return
+     */
+    @RequestMapping(value = "auth/indexLuceneWordNet", method = RequestMethod.POST, produces = "application/json")
+    public synchronized @ResponseBody IndexingEvent indexConcepts(HttpServletRequest req, Principal principal, ModelMap model) {
```
jdamerow: I wouldn't do this, this would create a bottle neck (all threads have to wait for the others to finish). It probably wouldn't be a big deal here (given that it's unlikely many people would start the indexer at the same time), but it's bad design and creates redundant calculations. Instead use a flag to check if the indexer is already running (e.g. AtomicBoolean) and only continue if it's not running. Otherwise tell the user that the indexer is currently running.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/lucene/impl/LuceneUtility.java
```diff
@@ -0,0 +1,442 @@
...
+                }
+            }
+        }
+
+        List<ConceptEntry> concepts = new ArrayList<ConceptEntry>();
+
+        try {
+            Query q = new QueryParser("", whiteSpaceAnalyzer).parse(queryString.toString());
+            DirectoryReader oldReader = reader;
+            reader = DirectoryReader.openIfChanged(oldReader);
```
jdamerow: Is there a reason that you reopen the reader here?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/web/LuceneIndexController.java
```diff
@@ -0,0 +1,97 @@
...
+        IndexingEvent bean = new IndexingEvent();
```
jdamerow: Instantiate this if you actually need it (catch block), otherwise it's just unnecessary use of memory/creation of objects. Same in the other methods.

jdamerow:
Please reopen this PR when you're done with the changes.

jdamerow:
Also, please make the date show up properly formatted using the users timezone. You can use simple Javascript for that.
image

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/lucene/impl/LuceneUtility.java
mkarthik90: @jdamerow Yes it is required because if there was an index run after the application has started and user tries to query after the run of index, the reader needs to be reloaded. When I don't reload it points to the old values. And also reader will be reloaded only if there is any change in the index, else reader will remain the same.

jdamerow: I would recreate the reader after you updated the index then, rather than do this if statement before every query. Conceptpower is more often queried that updated, so recreating the reader/searcher after an update makes more sense then doing it on every query.

PR #81, opened 2016-03-15
Ccp 128

mkarthik90:
[CCP-128][story/new-layout] Fixed review comments


Conceptpower+Spring/.settings/org.eclipse.m2e.core.prefs
```diff
@@ -1,4 +1,4 @@
-activeProfiles=dev
+activeProfiles=dev, dev
```
jdamerow: this seems redundant.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/web/LuceneIndexController.java
```diff
@@ -0,0 +1,106 @@
...
+     * @param req
+     * @param principal
+     * @param model
+     * @return
+     */
+    @RequestMapping(value = "auth/indexLuceneWordNet", method = RequestMethod.POST, produces = "application/json")
+    public @ResponseBody IndexingEvent indexConcepts(HttpServletRequest req, Principal principal, ModelMap model) {
+        IndexingEvent bean = null;
+        if (!indexerRunningFlag.get()) {
+            indexerRunningFlag.set(true);
```
jdamerow: This creates the same issues for thread safety as any other read/write action. What if a second thread gets to line 65 before the first thread move to line 66? You should use compareAndSet instead.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/lucene/impl/LuceneUtility.java
```diff
@@ -0,0 +1,442 @@
...
+
+        try {
+            Query q = new QueryParser("", whiteSpaceAnalyzer).parse(queryString.toString());
+            DirectoryReader oldReader = reader;
+            reader = DirectoryReader.openIfChanged(oldReader);
+            if (reader == null) {
+                reader = oldReader;
+            } else {
+                searcher = new IndexSearcher(reader);
+            }
```
jdamerow: see comment thread old PR

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/web/LuceneIndexController.java
jdamerow: also, this logic should be part of the indexservice not the controller. What if we allow to start the indexer from a rest interface? This implementation would work for that.

jdamerow:
see comments

PR #85, opened 2016-03-23
Ccp 135

mkarthik90:
[CCP-135] Fixed review comments and validated lucene exact search


Conceptpower+Spring/.classpath
```diff
@@ -33,5 +33,6 @@
 			<attribute name="maven.pomderived" value="true"/>
 		</attributes>
 	</classpathentry>
+	<classpathentry kind="lib" path="/Users/karthikeyanmohan/Downloads/art-uniroma2-lime-api-062ce14de8b4.zip"/>
```
jdamerow: please don't add library through your class path (should be in the pom) and don't commit this file either.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/lucene/impl/LuceneUtility.java
```diff
@@ -392,33 +393,23 @@ public void indexConcepts() throws LuceneException, IllegalArgumentException, Il
             if (search != null) {
```
jdamerow: I though this was already changes:
line 388/389 you have
```
ConceptEntry con = new ConceptEntry();
java.lang.reflect.Field[] fields = con.getClass().getDeclaredFields();
```
you don't have to create an object first, you can get the fields from the class itself.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/ConceptLookup.java
```diff
@@ -81,4 +92,25 @@
...
+        } catch (LuceneException ex) {
+            return new ResponseEntity<String>(ex.getMessage(), HttpStatus.BAD_REQUEST);
+        } catch (IllegalAccessException e) {
+            return new ResponseEntity<String>(e.getMessage(), HttpStatus.BAD_REQUEST);
+        }
+        Model model = ModelFactory.createDefaultModel();
+        String syntax = "RDF/XML";
+        StringWriter out = new StringWriter();
+        for (ConceptEntry entry : entries) {
+            rdfFactory.generateRDF(model, syntax, out, entry);
```
jdamerow: you probably don't want to have RDF classes in the controller. You should return a string from the RDF class or a custom object (if that works better), so that the controller doesn't have to be changed if the RDF library changes.

PR #87, opened 2016-03-25
Ccp 128

mkarthik90:
[CCP-128][story/new-layout] Changes for lucene and new layout for CCP


Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/core/impl/IndexService.java
```diff
@@ -0,0 +1,81 @@
...
+        luceneUtility.deleteIndexes();
+    }
+
+    /**
+     * This method indexes concepts in lucene and runs indexer only once
+     */
+    @Override
+    public String indexConcepts() throws LuceneException, IllegalArgumentException, IllegalAccessException {
+        if (!indexerRunningFlag.get()) {
+            indexerRunningFlag.compareAndSet(false, true);
```
jdamerow: you can use this method in the if statement. check the javadoc:
https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/atomic/AtomicInteger.html#compareAndSet(int,%20int)

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/core/impl/IndexService.java
```diff
@@ -0,0 +1,81 @@
...
+            luceneUtility.indexConcepts();
+            indexerRunningFlag.compareAndSet(true, false);
```
jdamerow: and here you then use simply "set" as you know what the value is at that point.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/core/impl/IndexService.java
```diff
@@ -0,0 +1,81 @@
...
+            return "Indexed successfully";
+        }
+        return "Indexer Already Running";
```
jdamerow: don't return a string. No string returns for determination of the the method is doing! Just return a boolean and your controller can then decide what to tell the user.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/core/impl/IndexService.java
```diff
@@ -0,0 +1,81 @@
...
+    public void deleteById(String id) throws LuceneException {
+        luceneUtility.deleteById(id);
+    }
+
+    /**
+     * This method deletes the index in lucene
+     */
+    @Override
+    public void deleteIndexes() throws LuceneException {
+        luceneUtility.deleteIndexes();
```
jdamerow: this needs to be guarded by the indexer running flag as well. otherwise I could delete your index while you are reindexing resulting in an half complete index.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/core/impl/IndexService.java
```diff
@@ -0,0 +1,81 @@
...
+        luceneUtility.insertConcept(entry);
+    }
+
+    /**
+     * This method deletes the concept in lucene index based on id of the
+     * concept
+     */
+    @Override
+    public void deleteById(String id) throws LuceneException {
+        luceneUtility.deleteById(id);
```
jdamerow: this needs to be guarded by the indexer running flag as well. otherwise I could delete or try to delete a concept that is already deleted (through the reindex/delete).

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/core/impl/IndexService.java
```diff
@@ -0,0 +1,81 @@
...
+            throws LuceneException, IllegalAccessException {
+        return luceneUtility.queryIndex(fieldMap, operator);
+    }
+
+    /**
+     * This method inserts concepts into lucene
+     */
+    @Override
+    public void insertConcept(ConceptEntry entry) throws IllegalAccessException, LuceneException {
+        luceneUtility.insertConcept(entry);
```
jdamerow: this needs to be guarded by the indexer running flag as well. or I could insert a new concept while someone else is reindexing which might result in the same concept inserted twice.

jdamerow: eg. someone is in the process of adding a new concept and hits add. at the same time someone hits reindex. the index gets deleted, the new concept is added, all concepts are indexed => new concept appears twice in index. make sense?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/core/impl/IndexService.java
```diff
@@ -0,0 +1,81 @@
...
+    private AtomicBoolean indexerRunningFlag = new AtomicBoolean(false);
+
+    /**
+     * This method searches in lucene on a particular condition fed through
+     * fieldMap
+     */
+    @Override
+    public ConceptEntry[] searchForConceptsConnected(Map<String, String> fieldMap, String operator)
+            throws LuceneException, IllegalAccessException {
+        return luceneUtility.queryIndex(fieldMap, operator);
```
jdamerow: I wouldn check the flag here, but before calling this method I would ask the idnexing service if a reindex is currently running, and if so tell the use to wait a minute until the indexer is done.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/web/LuceneIndexController.java
```diff
@@ -0,0 +1,97 @@
...
+    @Qualifier("luceneDAO")
+    private ILuceneDAO dao;
+
+    /**
+     * Retrives latest indexing time and number of concepts indexed so far
+     * 
+     * @param model
+     * @return
+     */
+    @RequestMapping(value = "auth/luceneIndex", method = RequestMethod.GET)
```
jdamerow: let's call this 'auth/index'. we don't want to change the URL if we ever want to change what indexer we use.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/web/LuceneIndexController.java
```diff
@@ -0,0 +1,97 @@
...
+    /**
+     * This method deletes all the index in Lucene and recreates all the index
+     * from Wordnet and from CCP db40 database
+     * 
+     * @param req
+     * @param principal
+     * @param model
+     * @return
+     */
+    @RequestMapping(value = "auth/indexLuceneWordNet", method = RequestMethod.POST, produces = "application/json")
```
jdamerow: and this maybe 'indexConcepts'?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/web/LuceneIndexController.java
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * This method deletes all the concepts in the lucene
+     * 
+     * @param req
+     * @param principal
+     * @param model
+     * @return
+     */
+    @RequestMapping(value = "auth/deleteConcepts", method = RequestMethod.POST, produces = "application/json")
```
jdamerow: auth/deleteIndex

mkarthik90:
@jdamerow Fixed the review comments. Let me know if you have any other comments

mkarthik90:
@jdamerow  resolved merge conflicts and fixed review comments

mkarthik90:
@jdamerow Added DS_STORE to gitignore and made changes to request mapping and method names in lucene controller.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/web/LuceneIndexController.java
```diff
@@ -0,0 +1,110 @@
...
+    @Qualifier("luceneDAO")
+    private ILuceneDAO dao;
+
+    /**
+     * Retrives latest indexing time and number of concepts indexed so far
+     * 
+     * @param model
+     * @return
+     */
+    @RequestMapping(value = "auth/loadIndex", method = RequestMethod.GET)
```
jdamerow: just call this "auth/index", load index makes it sound like it loads the index, but it really doesn't. It should shows you a page to reindex and with some info.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/web/LuceneIndexController.java
```diff
@@ -0,0 +1,110 @@
...
+     * @return
+     * @throws IndexerRunningException
+     */
+    @RequestMapping(value = "auth/indexConcepts", method = RequestMethod.POST, produces = "application/json")
+    public @ResponseBody IndexingEvent indexConcepts(HttpServletRequest req, Principal principal, ModelMap model)
+            throws IndexerRunningException {
+        IndexingEvent bean = null;
+        try {
+            if (manager.checkIndexerStatus()) {
+                bean = dao.getTotalNumberOfWordsIndexed();
```
jdamerow: you controller should only talk to the manager. No dependency on the DAO.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/core/impl/IndexService.java
```diff
@@ -0,0 +1,108 @@
...
+			throws LuceneException, IllegalArgumentException, IllegalAccessException, IndexerRunningException {
+		if (!indexerRunningFlag.compareAndSet(false, true)) {
+			throw new IndexerRunningException(ErrorConstants.INDEXER_RUNNING);
+		}
+		luceneUtility.indexConcepts();
+		indexerRunningFlag.set(false);
+	}
+
+	@Override
+	public boolean checkIndexerStatus() {
```
jdamerow: bad method name. Especially with boolean return values, the name should indicate how return value and method name relate, e.g. isWhatEverYouDoHappening -> yes/no -> true/false
in this case, e.g.: isIndexerRunning

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/core/ErrorConstants.java
```diff
@@ -0,0 +1,8 @@
+package edu.asu.conceptpower.servlet.core;
+
+public interface ErrorConstants {
+
+    public final static String INDEXERSTATUS = "IndexerStatus";
+    public final static String INDEXER_RUNNING = "Indexer running. Please try after sometime";
```
jdamerow: this is not a constant but an error message. This should be in a message file. Also, it should say: "The indexer is currently running. Please try again in a minute."

PR #92, opened 2016-04-06
Ccp 128

mkarthik90:
[CCP-128] Changes for review comments


jdamerow:
ok, i'll merge this now, so that we get it into develop, but there are bunch of things that still need to be fixed. I'll create extra tickets for those.

jdamerow:
ok to test

jdamerow:
Never mind, I won't merge it because there is a pretty big issue:
when you go to auth/user/list, you get this
image

Let's fix that first, and then I merge.

PR #98, opened 2016-04-15
Ccp 136

mkarthik90:
[CCP-136] Fixed review comments


mkarthik90:
Deleting the branch because it is stale and was updated a year ago.

PR #100, opened 2016-04-15
Changes for displaying error message while login fails

mkarthik90:
[CCP-140] Fixed error message display when login fails


jdamerow:
ok to test

PR #102, opened 2016-04-18
Ccp 143

mkarthik90:
[CCP-143] Changes for font and drop down styles


jdamerow:
ok to test

jdamerow:
There are spaces missing between the icons and text (in all places, not just the menu):
image

mkarthik90:
@jdamerow Fixed spacing issues

PR #105, opened 2016-05-12
[CCP-153] moved domain classes to old package structure

mkarthik90:
[CCP-153] Moved domain classes to old package structure


jdamerow:
hm, looks like you didn't commit the moved classes.

PR #125, opened 2016-09-07
[CCP-157] Changes for including uri in add concept

mkarthik90:
URI will be displayed in the output json when
concept is added through Rest calls


Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
```diff
@@ -98,7 +101,8 @@
             return new ResponseEntity<String>(jsonObject.toJSONString(), HttpStatus.CONFLICT);
```
jdamerow: see my previous PR comment. This should probably a different error code.

jdamerow:
ok to test

PR #126, opened 2016-09-08
[CCP-145] Overflow issue fixed in synonym modal

mkarthik90:
Synonym modal was overflowing when synonym
is searched in the modal. Fixed it by adding an
overflow auto and setting  a max-height


jdamerow:
ok to test

jdamerow:
there still seems to be something off:
image

mkarthik90:
@jdamerow  Fixed the issue. I missed few screens. Added the style for all screens.

jdamerow:
ok to test

PR #127, opened 2016-09-08
[CCP-141] Added a new end point creating concept wrappers

mkarthik90:
Users can use rest/conceptwrapper/add to add concept wrappers.
Mandatory fields for adding concept wrappers are concept list name
and word net ids.

### DIFF
--- a/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/ConceptWrapperRestController.java
+++ b/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/ConceptWrapperRestController.java
@@ -1,0 +1,57 @@
+package edu.asu.conceptpower.servlet.rest;
+
+import java.security.Principal;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.RestController;
+
+import edu.asu.conceptpower.core.ConceptEntry;
+import edu.asu.conceptpower.servlet.core.Constants;
+import edu.asu.conceptpower.servlet.core.IConceptManager;
+import edu.asu.conceptpower.servlet.core.IIndexService;
+import edu.asu.conceptpower.servlet.exceptions.DictionaryDoesNotExistException;
+import edu.asu.conceptpower.servlet.exceptions.DictionaryModifyException;
+import edu.asu.conceptpower.servlet.exceptions.IndexerRunningException;
+import edu.asu.conceptpower.servlet.exceptions.LuceneException;
+
+@RestController
+public class ConceptWrapperRestController {
+
+    @Autowired
+    private IConceptManager conceptManager;
+
+    @Autowired
+    private IIndexService indexService;
+
+    @RequestMapping(value = "rest/conceptwrapper/add", method = RequestMethod.POST)
+    public ResponseEntity<String> addConceptWrapper(@RequestBody ConceptEntry conceptEntry, Principal principal)
+            throws LuceneException, IllegalAccessException, DictionaryDoesNotExistException, DictionaryModifyException,
+            IndexerRunningException {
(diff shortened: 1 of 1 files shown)


Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/ConceptWrapperRestController.java
```diff
@@ -0,0 +1,57 @@
...
+public class ConceptWrapperRestController {
+
+    @Autowired
+    private IConceptManager conceptManager;
+
+    @Autowired
+    private IIndexService indexService;
+
+    @RequestMapping(value = "rest/conceptwrapper/add", method = RequestMethod.POST)
+    public ResponseEntity<String> addConceptWrapper(@RequestBody ConceptEntry conceptEntry, Principal principal)
```
jdamerow: this definitely needs javadoc that describes how this method can be used

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/ConceptWrapperRestController.java
```diff
@@ -0,0 +1,57 @@
...
+            throws LuceneException, IllegalAccessException, DictionaryDoesNotExistException, DictionaryModifyException,
+            IndexerRunningException {
+
+        if (conceptEntry.getConceptList() == null || conceptEntry.getConceptList().trim().isEmpty()) {
+            return new ResponseEntity<String>("Concept list cannot be empty", HttpStatus.BAD_REQUEST);
```
jdamerow: period missing at the end of the sentence.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/ConceptWrapperRestController.java
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        if (conceptEntry.getWordnetId() != null) {
+            String[] wrappers = conceptEntry.getWordnetId().split(Constants.CONCEPT_SEPARATOR);
+            if (wrappers.length > 0) {
+                conceptEntry.setWord(conceptManager.getConceptEntry(wrappers[0]).getWord().replace("_", " "));
+                conceptEntry.setPos(conceptManager.getConceptEntry(wrappers[0]).getPos());
```
jdamerow: instead of calling the database twice, just do it once and use the concept entry you get.

also, if the id doesn't exist you will get a null pointer here. a non-existing wordnet id should be handles appropriately.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/ConceptWrapperRestController.java
```diff
@@ -0,0 +1,57 @@
...
+                conceptEntry.setCreatorId(principal.getName());
+
+                if (indexService.isIndexerRunning()) {
+                    return new ResponseEntity<String>("Indexer running. Try again after some time.",
+                            HttpStatus.INTERNAL_SERVER_ERROR);
```
jdamerow: this error code should be consistent with the other rest methods.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/ConceptWrapperRestController.java
```diff
@@ -0,0 +1,57 @@
...
+                }
+                conceptManager.addConceptListEntry(conceptEntry, principal.getName());
+            }
```
jdamerow: what about a type, synonyms, equal to, description, etc for a concept? this method should accept almost the same fields as the regular add method. You might want to think about adding the wrapper functionality to the regular add method, or at least reuse code from there.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/ConceptWrapperRestController.java
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        return new ResponseEntity<String>("Concept wrapper created successfully", HttpStatus.OK);
```
jdamerow: missing period

PR #128, opened 2016-09-08
[CCP-164] Details are parsed now to be displayed

mkarthik90:
Details from ajax call is parsed and displayed on the
modal.


jdamerow:
ok to test

PR #130, opened 2016-09-13
CCP-141 Added rest call for creating concept wrappers

mkarthik90:


Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/core/IConceptManager.java
```diff
@@ -20,7 +20,7 @@
 	 *            of entry
 	 * @return Entry for ID or null.
 	 */
-	public abstract ConceptEntry getConceptEntry(String id)throws LuceneException;
+    public abstract ConceptEntry getConceptEntry(String id);
```
mkarthik90: Removed LuceneException since this method is not throwing any lucene exception

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
```diff
@@ -68,7 +87,14 @@
                     HttpStatus.BAD_REQUEST);
         }
 
-        JsonValidationResult result = checkJsonObject(jsonObject);
+        JsonValidationResult result = null;
+        if (jsonObject.get(JsonFields.WORDNET_ID) != null) {
+            result = checkJsonObjectForWrapper(jsonObject);
+        }
+        else {
```
jdamerow: formatting

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
```diff
@@ -47,6 +48,24 @@
...
+     * wrapper ids then concept is wrapped else the new concept is created.
+     * 
+     * Sample input
+     * 
+     * { "wordnetids":
+     * "WID-00450866-N-01-pony-trekking,WID-03981924-N-01-pony_cart",
+     * "synonymids" : " WID-02380464-N-01-polo_pony,WID-04206225-N-03-pony",
+     * "conceptlist" : "FirstList-1", "type" :
+     * "88bea1dc-1443-4296-8315-715c71128b01" "description" : "Description",
+     * "equalsTo" : "equals", "similarTo" : "similar" }
```
jdamerow: this should have also a json example for regular adding a concept and make clear what json does what.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
```diff
@@ -189,10 +222,21 @@ private JsonValidationResult checkJsonObject(JSONObject jsonObject) {
                     false);
         }
 
-        if (jsonObject.get(JsonFields.DESCRIPTION) == null) {
+        String pos = jsonObject.get(JsonFields.POS).toString();
```
jdamerow: if POS is not defined this will probably throw a nullpointer

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
```diff
@@ -189,10 +222,21 @@ private JsonValidationResult checkJsonObject(JSONObject jsonObject) {
                     false);
         }
 
-        if (jsonObject.get(JsonFields.DESCRIPTION) == null) {
```
jdamerow: did we decide not to make description required?

jdamerow: never mind. overlooked the other call at the end.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
```diff
@@ -201,35 +245,40 @@ private JsonValidationResult checkJsonObject(JSONObject jsonObject) {
...
         }
-
-        String pos = jsonObject.get(JsonFields.POS).toString();
-        if (!POS.posValues.contains(pos)) {
-            logger.error("Error creating concept from REST call. " + pos + " does not exist.");
-            return new JsonValidationResult("POS '" + pos + "' does not exist.", jsonObject, false);
-        }
-
         return new JsonValidationResult(null, jsonObject, true);
     }
```
jdamerow: if the user submits a concept wrapper with pos = noun, but the wrapped wordnet concept is pos = verb, there should be a validation error raised.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
jdamerow: never mind it, it's tested in the beginning of the method. but you should keep things together that belong together.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
```diff
@@ -189,10 +222,21 @@ private JsonValidationResult checkJsonObject(JSONObject jsonObject) {
                     false);
         }
 
-        if (jsonObject.get(JsonFields.DESCRIPTION) == null) {
+        String pos = jsonObject.get(JsonFields.POS).toString();
+        if (!POS.posValues.contains(pos)) {
+            logger.error("Error creating concept from REST call. " + pos + " does not exist.");
+            return new JsonValidationResult("POS '" + pos + "' does not exist.", jsonObject, false);
```
jdamerow: no need to log here, we don't log the other ones. Also, it's definitely not an error that needs to be logged, if anything than maybe debug. 

For the error message, please follow the same format that the other validation messages use.

jdamerow: ah, that was already in there. but please correct it anyways ;)

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
```diff
@@ -201,35 +245,40 @@ private JsonValidationResult checkJsonObject(JSONObject jsonObject) {
...
+
     private ConceptEntry createEntry(JSONObject jsonObject, String username) {
         ConceptEntry conceptEntry = new ConceptEntry();
-        conceptEntry.setCreatorId(username);
-        conceptEntry.setSynonymIds(jsonObject.get(JsonFields.SYNONYM_IDS) != null ? jsonObject.get(
-                JsonFields.SYNONYM_IDS).toString() : "");
-        conceptEntry.setWord(jsonObject.get(JsonFields.WORD).toString());
+        if (jsonObject.get(JsonFields.WORDNET_ID) != null) {
+            conceptEntry.setWordnetId(jsonObject.get(JsonFields.WORDNET_ID).toString());
+            String[] wrappers = jsonObject.get(JsonFields.WORDNET_ID).toString().split(Constants.CONCEPT_SEPARATOR);
```
jdamerow: I've seen that in multiple spots but now I'm thinking about hat. Why does this field expects to get a comma separated list? So far, there can only be one wordnet id, right?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
```diff
@@ -201,35 +245,40 @@ private JsonValidationResult checkJsonObject(JSONObject jsonObject) {
...
+            if (wrappers.length > 0) {
+                ConceptEntry existingConceptEntry = conceptManager.getConceptEntry(wrappers[0]);
+                conceptEntry.setWord(existingConceptEntry.getWord().replace("_", " "));
```
jdamerow: if you always use the word from wordnet here, you should validate that the user doesn't submit a word, or make it clear in the javadoc that any submitted word is being ignored.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/JsonFields.java
```diff
@@ -19,4 +19,5 @@
     public final static String SIMILAR = "similar";
     public final static String TYPE = "type";
     public final static String URI = "uri";
+    public final static String WORDNET_ID = "wordnetids";
```
jdamerow: camel case: capitalize 'i' of 'ids'

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
mkarthik90: I checked in UI. In create concept wrapper page we are able to add more than one wordnet id. And the wrapper is created for both of those. I have attached a sample screenshot for a wrapper with two wordnet ids. 
screen shot 2016-09-14 at 11 01 16 am

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
mkarthik90: @jdamerow I think we are not accepting pos for concept wrappers. We are fetching the pos and word details from wordnet. I have included this in the note in javadocs. Let me know if we should allow the user to submit the pos details and we need to validate and throw the error you have mentioned.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
mkarthik90: Made it as debug

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
mkarthik90: Included a note saying we will ignore word and pos for concept wrappers.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
jdamerow: when you not accept a value then there should be an error, otherwise you're accepting it. Overriding is fine for the 'word' property if it's mentioned in the javadoc, but I would actually returning an error when the user submits a POS different than the one in wordnet, since that might mean the user thought they selected a different wordnet.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
jdamerow: Hm, ok, but if we allow this, then the validation needs to check that accordingly.

mkarthik90: Validated wordnet ids.

mkarthik90:
@jdamerow Added validations for wordnets and pos.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
```diff
@@ -201,38 +265,74 @@ private JsonValidationResult checkJsonObject(JSONObject jsonObject) {
...
+                // In case user has entered a POS value. Validate whether POS is
+                // same as wordnet POS
+                if (jsonObject.get(JsonFields.POS) != null) {
+                    if (!existingConceptEntry.getPos().equalsIgnoreCase(jsonObject.get(JsonFields.POS).toString())) {
+                        throw new POSMismatchException("Entered POS " + jsonObject.get(JsonFields.POS).toString()
```
jdamerow: You should validate that POSs are correct in the validation function and return a proper validation method like all the other validations. Exception should be used for unanticipated situations, which isn't really the case here.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
```diff
@@ -201,38 +265,74 @@ private JsonValidationResult checkJsonObject(JSONObject jsonObject) {
...
+                jsonObject.get(JsonFields.SIMILAR) != null ? jsonObject.get(JsonFields.SIMILAR).toString() : "");
         conceptEntry.setTypeId(jsonObject.get(JsonFields.TYPE).toString());
+
         return conceptEntry;
     }
 
+    @ExceptionHandler({ POSMismatchException.class })
+    public ResponseEntity<String> posMismatchHandler(HttpServletRequest req, Exception ex) {
+        return new ResponseEntity<String>(ex.getLocalizedMessage(), HttpStatus.BAD_REQUEST);
+    }
```
jdamerow: why would you use a different technique just for this one validation?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
```diff
@@ -201,38 +265,74 @@ private JsonValidationResult checkJsonObject(JSONObject jsonObject) {
...
+            String wordnetIds = jsonObject.get(JsonFields.WORDNET_ID).toString();
+            List<String> wordnetIdsList = Arrays
+                    .asList(wordnetIds.split("\\s*" + Constants.CONCEPT_SEPARATOR + "\\s*"));
+            for (String wordNetId : wordnetIdsList) {
+                try{ 
+                    WordID.parseWordID(wordNetId);
+                } catch (IllegalArgumentException ex) {
+                    return new JsonValidationResult(
+                            "Error parsing request: please provide a valid list of wordnet ids seperated by commas.",
+                            jsonObject, false);
```
jdamerow: this should say a little more. What wordnet id does not exist?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
```diff
@@ -130,7 +176,14 @@
         while (listIt.hasNext()) {
             JSONObject jsonObject = listIt.next();
 
-            JsonValidationResult result = checkJsonObject(jsonObject);
+            JsonValidationResult result = null;
+
+            if (jsonObject.get(JsonFields.WORDNET_ID) != null) {
+                result = checkJsonObjectForWrapper(jsonObject);
+            }
+            else {
```
jdamerow: formatting

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
```diff
@@ -47,9 +55,40 @@
 
     private static final Logger logger = LoggerFactory.getLogger(Concepts.class);
 
+    /**
+     * This method creates concepts as well as concept wrappers. This is decided
+     * based on the request parameter. If the request parameter contains the
+     * wrapper ids then concept is wrapped else the new concept is created.
```
jdamerow: you probably mean "contains one or more Wordnet Ids"

jdamerow: And also "then a new concept wrapper is created, otherwise a new concept is created."

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
```diff
@@ -47,9 +55,40 @@
...
+     * 
+     * Sample input for creating a concept wrapper:
+     * 
+     * { "wordnetIds":
+     * "WID-00450866-N-01-pony-trekking,WID-03981924-N-01-pony_cart",
+     * "synonymids" : " WID-02380464-N-01-polo_pony,WID-04206225-N-03-pony",
+     * "conceptlist" : "FirstList-1", "type" :
+     * "88bea1dc-1443-4296-8315-715c71128b01", "description" : "Description",
+     * "equalsTo" : "equals", "similarTo" : "similar" }
```
jdamerow: Please format that so that it's easily readable.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
```diff
@@ -47,9 +55,40 @@
...
+     * Please note any word or pos passed during concept wrapper creation will
+     * be ignored. Word and pos details are fetched from wordnet. In case POS is
+     * entered by user, the code validates with the wordnet POS and throws an
+     * error if POS is not matching with wordnet
+     * 
+     * Sample input for creating a concept:
+     * 
+     * { "word":"kitty", "pos": "noun", "conceptlist":"mylist", "description":
+     * "Soft kitty, sleepy kitty, little ball of fur.",
+     * "type":"3b755111-545a-4c1c-929c-a2c0d4c3913b" }
```
jdamerow: that too

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
```diff
@@ -183,16 +236,27 @@ private JsonValidationResult checkJsonObject(JSONObject jsonObject) {
                     jsonObject, false);
         }
 
+        String pos = jsonObject.get(JsonFields.POS).toString();
+        if (!POS.posValues.contains(pos)) {
+            logger.debug("Error creating concept from REST call. " + pos + " does not exist.");
+            return new JsonValidationResult("POS '" + pos + "' does not exist.", jsonObject, false);
```
jdamerow: Please format the validation message like all the other sones.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
mkarthik90: @jdamerow  I can include the validation in checkJsonObjectForWrapper() but there is a problem

1) We need to hit the db and get the existing pos value. We are again hitting the db in createEntry method.

So I thought I can hit the db once and validate that time. Let me know if we can do it better or is it ok to hit db twice

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
mkarthik90: If I include the validation check in checkJsonObjectForWrapper() I can do it as in the existing

mkarthik90:
@jdamerow  Fixed review comments. Pls let me know if i need to change the validation of POS

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
jdamerow: I doubt that this would be a huge deal as so far don't really submit many concepts as one, so if you make one or two calls to the db doesn't really affect anything. But if you're worried about that, why don't you just retrieve the wrapped concept (if there is any) before you call validation or creation and just pass it to the to method. This way you would avoid retrieving it twice.

mkarthik90:
@jdamerow  Fixed review comments. Removed POSMismatchException and handled this validation as per other validations

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
```diff
@@ -201,35 +267,67 @@ private JsonValidationResult checkJsonObject(JSONObject jsonObject) {
...
+            String wordnetIds = jsonObject.get(JsonFields.WORDNET_ID).toString();
+            List<String> wordnetIdsList = Arrays
+                    .asList(wordnetIds.split("\\s*" + Constants.CONCEPT_SEPARATOR + "\\s*"));
+            for (String wordNetId : wordnetIdsList) {
+                try{ 
+                    WordID.parseWordID(wordNetId);
+                } catch (IllegalArgumentException ex) {
+                    return new JsonValidationResult(
+                            "Error parsing request: please provide a valid list of wordnet ids seperated by commas. Wordnet id "
+                                    + wordNetId + " is not matching.",
```
jdamerow: matching what? should be wordnet id "does not exist."

mkarthik90:
@jdamerow Fixed review comments

jdamerow:
ok to test

jdamerow:
Submitting the following json causes a Nullpointer:
```
{
	"word": "Sauce Albert",
	"pos": "verb",
	"description": "Sauce Albert yammy",
	"coceptlist": "test",
	"type": "88bea1dc-1443-4296-8315-715c71128b01",
	"wordnetIds": "WID-07830841-N-02-sauce_Albert"
}
```

The validation for a missing concept list does not seem to work.

jdamerow:
After adding a wrapper, the search is somehow broken for that term: 
- before I added a wrapper for "sauce albert", the term showed up when I searched for "albert" (you can search on production and you'll see it shows up)
- after I added the wrapper, "sauce albert" now doesn't show up in the search results anymore. I have to explicitly search for "sauce albert"

jdamerow:
Also, it should not be possible to create two wrappers for the same concept. Once a wordnet concept is wrapped, it can't be wrapped again.

mkarthik90:
@jdamerow  
1) The conceptlist has a spelling error, thats why validation throws error.
2) I tested for a valid type id in local region, the add was successful. 
3) After adding the same concept wrapper in the local region I was able to search it through the home screen successfully. Did you try searching with a verb POS. Because the pos of "sauce albert" by default was Noun but we have added as a verb. Should we have a validation to check if the wrapped concept is also of same pos type? 
screen shot 2016-10-11 at 8 35 50 am

mkarthik90:

1) Search results will show wrapped concept. 
2) User can wrap the wordnet ids only once.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/core/impl/ConceptManager.java
```diff
@@ -472,4 +473,21 @@ public void deleteConcept(String id, String userName) throws LuceneException, In
         client.update(concept, DBNames.DICTIONARY_DB);
         indexService.deleteById(concept.getId());
     }
+
+    @Override
+    public ConceptEntry getConceptWrappedEntryByWordNetId(String wordNetID)
```
jdamerow: why don't you ask the database directly?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/Concepts.java
```diff
@@ -201,35 +274,77 @@ private JsonValidationResult checkJsonObject(JSONObject jsonObject) {
...
+                            jsonObject, false);
+                }
+            }
+
+            // Check if there is an existing wrapper concept if so throw error
+            for (String wordNetId : wordnetIdsList) {
+                if (conceptManager.getConceptWrappedEntryByWordNetId(wordNetId) != null) {
+                    // This wordnet id is already wrapped. Throw error
+                    return new JsonValidationResult(
+                            "Error parsing request: please provide a valid wordnet id that doesn't have a wrapped concept.",
```
jdamerow: should say:
"...: the WordNet concept you are trying to wrap is already wrapped."

mkarthik90:

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/core/impl/ConceptManager.java
```diff
@@ -477,10 +476,8 @@ public void deleteConcept(String id, String userName) throws LuceneException, In
...
             throws IllegalAccessException, LuceneException, IndexerRunningException {
-        Map<String, String> fieldMap = new HashMap<>();
-        fieldMap.put(LuceneFieldNames.WORDNETID, wordNetID);
-        ConceptEntry[] entries = indexService.searchForConcepts(fieldMap, SearchParamters.OP_AND);
-        for (ConceptEntry entry : entries) {
+        List<ConceptEntry> conceptEntries = client.getConceptByWordnetId(wordNetID);
+        for (ConceptEntry entry : conceptEntries) {
             // Wordnet is also added because lucene doesn't do an exact search
             // on fields
             if (entry.getId().contains("CON") && entry.getWordnetId().contains(wordNetID)) {
```
jdamerow: Since you're asking for concepts by their concept id, isn't this always true:
```
entry.getWordnetId().contains(wordNetID)
```

jdamerow: Also, the id should start with "CON" and isn't there a method that returns "CON" from where the ids are created? This is not very good design. What if we change how the ids are created again?

jdamerow:
please merge conflicts

mkarthik90:
@jdamerow  Merged with develop and fixed review comments

jdamerow:
I can create wordnet wrapper twice for the same wordnet id. I created two water polo concepts for wordnet id: WID-00464478-N-01-water_polo

mkarthik90:
@jdamerow  There was an issue while check for wordnet ids in database. I fixed that and reopened the PR. Pls validate now.

jdamerow:
ok to test

jdamerow:
After creating a wrapper, the wrapped concept doesn't who up in the search anymore.

mkarthik90:
@jdamerow  I did something as below. 

First I created a wrapper for **WID-02382635-N-01-Shetland_pony** I created with the details as shown in the image. 

screen shot 2016-10-25 at 11 09 39 am


I then searched for the word pony from home screen search. The attached concept wrapper showed up. 

screen shot 2016-10-25 at 11 09 31 am


Please let me know if I am doing something wrong

jdamerow:
Hm, weird. Don't know what happened there. Please reformat IConceptManager.java and then I'll merge.

mkarthik90:
@jdamerow Addressed review comments. Reformatted IConceptManager

PR #131, opened 2016-09-15
[CCP-168] Form will not be submitted when enter is pressed

mkarthik90:
Following changes are made to concept add screen.
1) When user does not enter any value for Term field
and tries to click on search, we do not hit the database,
instead we display error message.
2) In case of any error during ajax call we display
processing error to the user.
3) When user tries to search a word that is not in VIAF an
unknown alert box was appearing on the screen. Now handled
that condition by clearing the data table and displaying proper error
message.
4) Changed the button size to small to make it look better.

### DIFF
--- a/Conceptpower+Spring/src/main/webapp/WEB-INF/views/auth/concepts/ConceptAddView.jsp
+++ b/Conceptpower+Spring/src/main/webapp/WEB-INF/views/auth/concepts/ConceptAddView.jsp
@@ -247,6 +247,13 @@
     $(function() {
         $("#serviceSearch").click(function() {
             var serviceterm = $("#serviceterm").val();
+            // Added validation for search term
+            if(!serviceterm.trim()) {
+                $("#serviceTermError").html('Term cannot be empty.');
+                return false;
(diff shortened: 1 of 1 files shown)


jdamerow:
ok to test

jdamerow:
retest this please

jdamerow:
Hm, hitting enter in the search box does still not search.

mkarthik90:
@jdamerow  Added changes to make sure when enter is pressed on the search field, the term search happens. And when enter is pressed on other fields entire form gets submitted.

jdamerow:
ok to test

jdamerow:
Since we're at it: when I hit enter in any other field the form gets submitted. If there are validation errors, the spinning icon for the search on top is continuously shown and spinning (which shouldn't be the case):
image

mkarthik90:
screen shot 2016-09-21 at 11 39 31 am

jdamerow:
ok to test

mkarthik90:
@jdamerow  Fixed review comments

PR #132, opened 2016-09-15
[CCP-169] Added encoding for characters for all responses

mkarthik90:
Spring by default uses ISO-8859-1 so characters were appearing
as ?? in UI. In UI we are using UTF-8. Added message converters
to make sure characters are encoded in UTF-8 format.


mkarthik90:
Sample screenshot after adding UTF-8. The characters looks fine. 

screen shot 2016-09-15 at 1 06 31 pm

jdamerow:
ok to test

PR #133, opened 2016-09-15
[CCP-170] Description will not be escaped anymore

mkarthik90:
Description was not appearing properly in UI. HTML tags
were escaped and rendered as normal string. From now on
all html tags will be rendered properly for description alone.


jdamerow:
ok to test

jdamerow:
I made a fix for this a while ago as hotfix to master, but the changes must have been lost when the classes got moved. I also fixed the leading line break in descriptions of wordnet wrappers. Please apply that fix here as well:
https://github.com/diging/conceptpower/blob/026e1b4de1d939c6e5de1bbd3be53a4e7817b7c3/Conceptpower%2BSpring/src/main/java/edu/asu/conceptpower/wrapper/impl/ConceptEntryWrapperCreator.java

mkarthik90:
@jdamerow  I checked the blame for ConceptEntryWrapperCreator. I am not able to find any changes on description. Is there any way I can check that.

mkarthik90:
@jdamerow Added the description changes

jdamerow:
ok to test

jdamerow:
description is empty now:
image

jdamerow:
ok to test

PR #134, opened 2016-09-22
CCP-172 Pagination changes

mkarthik90:


Conceptpower+Spring/src/main/java/edu/asu/conceptpower/root/DatabaseClient.java
```diff
@@ -296,6 +302,84 @@ public boolean match(ConceptList dict) {
      * lang.String)
      */
     @Override
+    public List<ConceptEntry> getAllEntriesFromList(String listname, int page, int pageSize, final String sortBy,
+            final int sortDirection) {
+        ConceptEntry entry = new ConceptEntry();
+        entry.setConceptList(listname);
+
+        Query wordNetQuery = wordnetCacheClient.query();
```
jdamerow: why do you query the wordnet cache if you get all entries in a list here? wordnet concepts are by default in the wordnet list and this list is not accessible through the concept list webpage.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/root/DatabaseClient.java
```diff
@@ -296,6 +302,84 @@ public boolean match(ConceptList dict) {
...
+                        } else {
+                            o2FieldContent = sortField.get(o1);
+                            o1FieldContent = sortField.get(o2);
+                        }
+                    } catch (IllegalArgumentException | IllegalAccessException e) {
+                        logger.error("Error accessing field.", e);
+                        return 0;
+                    }
+
+                    if (sortBy.endsWith("Date")) {
```
jdamerow: I don't think that the date fields end with 'Date' in Conceptpower? Also, since it's using change events, this is kind of obsolete I think.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/web/ConceptDeleteController.java
```diff
@@ -139,7 +143,7 @@ public ModelAndView confirmlDelete(@PathVariable("id") String id,
             model.setViewName("redirect:/login");
             return model;
         }
-        model.setViewName("/auth/conceptlist/concepts");
+        model.setViewName("forward:/auth/" + listName + "/concepts");
```
jdamerow: if you're forwarding here, you don't need to retrieve the concepts first (line 135-139), do you?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/web/ConcepTypeDeleteController.java
```diff
@@ -63,7 +64,8 @@ public String prepareDeleteType(@PathVariable("typeid") String typeid, ModelMap
         boolean enableDelete = true;
         List<ConceptList> conceptLists = conceptListManager.getAllConceptLists();
         for (ConceptList conceptList : conceptLists) {
-            List<ConceptEntry> conceptEntries = conceptManager.getConceptListEntries(conceptList.getConceptListName());
+            List<ConceptEntry> conceptEntries = conceptManager.getConceptListEntries(conceptList.getConceptListName(),
+                    1, -1, "id", IConceptDBManager.DESCENDING);
```
jdamerow: if i'm not mistaken,this will break the delete. It's supposed to check all concept for the given type, but you retrieve only the first page, don't you? 

Checking all concept should be done differently (asking directly the database for concept with a specific type) independently of your code, but I think the way it is right will break it.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/root/DatabaseClient.java
mkarthik90: This method was fetching the concept entries from both wordnetcache client and dictionary cache client and grouped it together. So I followed the same approach of fetching the concept entries from wordnet cache and dictionary cache. Please let me know if I need to remove wordnet cache client here.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/root/DatabaseClient.java
mkarthik90: Removed it

mkarthik90:
@jdamerow  Fixed review comments

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/root/DatabaseClient.java
jdamerow: What I'm saying is that this method ```getAllEntiresFromList``` is supposed to return all the entries of a concept list (now paginated), right? The entries that this method return can never be in the wordnet cache, because only concepts not in wordnet (or wrapped wordnet concepts) will be in a concept list and hence they will be stored in the concept database (not the wordnet cache). So, any concept in the wordnet cache simply cannot be in a concept list and therefore cannot be retrieved by this method. Does this make sense?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/web/ConceptDeleteController.java
```diff
@@ -129,17 +132,17 @@ public ModelAndView confirmlDelete(@PathVariable("id") String id,
...
-
-        model.addObject("result", foundConcepts);
         if (fromHomeScreenDelete.equalsIgnoreCase("true")) {
+            List<ConceptEntry> founds = conceptManager.getConceptListEntries(concept.getConceptList(), 1, -1, "id",
+                    IConceptDBManager.DESCENDING);
+
+            foundConcepts = wrapperCreator.createWrappers(
+                    founds != null ? founds.toArray(new ConceptEntry[founds.size()]) : new ConceptEntry[0]);
+            model.addObject("result", foundConcepts);
             model.setViewName("redirect:/login");
```
jdamerow: if you're redirecting here, I don't think the results will be forwarded to the redirect, will they? You probably have to add them as redirect attributes.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/web/ConceptListEditController.java
```diff
@@ -115,7 +116,8 @@ public String editList(HttpServletRequest req,
 
         // modify the name for all the existing concepts under this concept
         // list
-        List<ConceptEntry> entries = conceptManager.getConceptListEntries(conceptListAddForm.getOldListName());
+        List<ConceptEntry> entries = conceptManager.getConceptListEntries(conceptListAddForm.getOldListName(), 1, -1,
+                "id", IConceptDBManager.DESCENDING);
```
jdamerow: I think this is the same issue as for deleting a concept type. Here you are only getting the first page of concepts and change the concept list. For all other concepts in the list, the concept list name doesn't change.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/root/DatabaseClient.java
mkarthik90: yeah got it now. I will remove the wordnet cache then

mkarthik90:

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/root/DatabaseClient.java
```diff
@@ -309,6 +377,20 @@ public boolean match(ConceptList dict) {
         return allResults;
     }
 
+    @Override
+    public List<ConceptEntry> getAllEntriesByTypeId(String typeId) {
+        ConceptEntry entry = new ConceptEntry();
+        entry.setTypeId(typeId);
+
+        List<ConceptEntry> results = wordnetCacheClient.queryByExample(entry);
```
jdamerow: looking at the class, it seems like the wordnetcache is not used (I originally used it to improve the search, but now we have lucene, so this is not needed anymore). The code that would store concepts in the cache is commented out. This line would therefore never return any results. You should be able to remove any queries to the wordnet cache.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/web/ConcepTypeDeleteController.java
```diff
@@ -61,13 +60,10 @@ public String prepareDeleteType(@PathVariable("typeid") String typeid, ModelMap
...
-        List<ConceptList> conceptLists = conceptListManager.getAllConceptLists();
-        for (ConceptList conceptList : conceptLists) {
-            List<ConceptEntry> conceptEntries = conceptManager.getConceptListEntries(conceptList.getConceptListName());
-            for (ConceptEntry conceptEntry : conceptEntries) {
-                if ((conceptEntry.getTypeId() != null) && (conceptEntry.getTypeId()).equals(typeid)) {
-                    enableDelete = false;
-                }
+        List<ConceptEntry> conceptEntries = conceptManager.getConceptEntryByTypeId(typeid);
+        for (ConceptEntry conceptEntry : conceptEntries) {
+            if ((conceptEntry.getTypeId() != null) && (conceptEntry.getTypeId()).equals(typeid)) {
```
jdamerow: since you're requesting concepts that have the type=typeid, isn't this always true?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/web/ConceptDeleteController.java
```diff
@@ -114,7 +117,9 @@ public String cancelDelete(@PathVariable("conceptList") String conceptList, Mode
      */
     @RequestMapping(value = "auth/conceptlist/deleteconceptconfirm/{id}", method = RequestMethod.GET)
     public ModelAndView confirmlDelete(@PathVariable("id") String id,
```
jdamerow: this should be a POST request.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/web/ConceptDeleteController.java
jdamerow: I think 'redirect' is the better way to go here. Otherwise you're browser will still show the previous url and a reload of the page would call the delete method again.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/core/IConceptManager.java
```diff
@@ -118,5 +121,23 @@ public abstract void storeModifiedConcept(ConceptEntry entry, String userName)
...
+     */
+    public List<ConceptEntry> getConceptEntryByTypeId(String typeId);
+
+    /**
+     * Gets all the concept entry for the specified concept list name
+     * 
+     * @param conceptListName
+     * @return
+     */
+    public List<ConceptEntry> getConceptEntriedByConceptListName(String conceptListName);
```
jdamerow: why don't you overload the ```getConceptListEntries``` method as you do in the database client?

mkarthik90:

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/web/ConcepTypeDeleteController.java
```diff
@@ -62,7 +54,7 @@ public String prepareDeleteType(@PathVariable("typeid") String typeid, ModelMap
         boolean enableDelete = true;
         List<ConceptEntry> conceptEntries = conceptManager.getConceptEntryByTypeId(typeid);
         for (ConceptEntry conceptEntry : conceptEntries) {
-            if ((conceptEntry.getTypeId() != null) && (conceptEntry.getTypeId()).equals(typeid)) {
+            if (conceptEntry.getTypeId() != null) {
```
jdamerow: since you've asked for only concepts with a specific type, can the type ever be null?

Conceptpower+Spring/src/main/webapp/WEB-INF/views/auth/concepts/ConceptDeleteView.jsp
```diff
@@ -13,70 +13,70 @@
...
-		<c:if test="${fromHomeScreenDelete eq 'true'}">
-		<td><a
-			href="${pageContext.servletContext.contextPath}/login"><input
-				type="button" name="cancel" value="No, cancel!" class="btn btn-primary"></a></td>
-		</c:if>
-		
-	</tr>
-</table>
\ No newline at end of file
+<form:form id="deleteConfirmForm" action="${pageContext.servletContext.contextPath}/auth/conceptlist/deleteconceptconfirm/${conceptId}?fromHomeScreenDelete=${fromHomeScreenDelete}&listName=${conceptList}">
```
jdamerow: does this work? don't you have to specify that this should be a POST request?

mkarthik90:

jdamerow:
ok to test

jdamerow:
Cool. Good work.

PR #137, opened 2016-10-18
CCP-176 Pagination using lucene for conceptsearch

mkarthik90:


mkarthik90:
@jdamerow  Added the changes for CCP-175

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/core/impl/IndexService.java
```diff
@@ -53,10 +53,26 @@
 		if (indexerRunningFlag.get()) {
 			throw new IndexerRunningException(indexerRunning);
 		}
-
-		return luceneUtility.queryIndex(fieldMap, operator);
+        // Fetches all the pages
+        return luceneUtility.queryIndex(fieldMap, operator, Integer.MIN_VALUE, Integer.MIN_VALUE);
```
jdamerow: why don't you put in 0 for start index and -1 for page number? and then have logic in your lucene utility that checks for -1? This seems odd and not very intuitive.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/ConceptSearch.java
```diff
@@ -55,29 +60,53 @@
...
      *            Holds HTTP request information
      * @return
+     * @throws IllegalAccessException
+     * @throws IllegalArgumentException
      */
     @RequestMapping(value = "rest/ConceptSearch", method = RequestMethod.GET, produces = "application/xml; charset=utf-8")
-    public @ResponseBody ResponseEntity<String> searchConcept(HttpServletRequest req) {
-        Map<String, String[]> queryParams = req.getParameterMap();
+    public @ResponseBody ResponseEntity<String> searchConcept(
+            @Valid ConceptSearchParameters conceptSearchParameters)
```
jdamerow: It's a good idea to validate those parameters, but there doesn't seem to be a validator?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/ConceptSearch.java
```diff
@@ -55,29 +60,53 @@
...
         ConceptEntry[] searchResults = null;
 
+        int totalNumberOfRecords = 0;
         try {
-            searchResults = manager.searchForConcepts(searchFields, operator);
+            totalNumberOfRecords = manager.getTotalNumberOfRecordsForSearch(searchFields, operator);
+            searchResults = manager.searchForConceptByPageNumberAndFieldMap(searchFields, operator,
+                    page, numberOfRecordsPerPage);
         } catch (LuceneException | IllegalAccessException | IndexerRunningException ex) {
             return new ResponseEntity<String>(ex.getMessage(), HttpStatus.BAD_REQUEST);
```
jdamerow: Please respond with correct error codes here.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/ConceptSearch.java
```diff
@@ -55,29 +60,53 @@
...
         }
 
+        if (searchResults.length == 0) {
+            // Returning a 404 error.
+            return new ResponseEntity<String>("No records found for the search condition.",
+                    HttpStatus.NOT_FOUND);
```
jdamerow: shouldn't be not found. Not found would mean the resource wasn't found, but it was. It just didn't yield any results, which is just as ok as results.

mkarthik90:

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/ConceptSearch.java
```diff
@@ -97,14 +99,16 @@
...
+        } catch (LuceneException | IllegalAccessException ex) {
+            return new ResponseEntity<String>(ex.getMessage(), HttpStatus.INTERNAL_SERVER_ERROR);
+        } catch (IndexerRunningException e) {
+            return new ResponseEntity<String>(e.getMessage(), HttpStatus.SERVICE_UNAVAILABLE);
         }
 
         if (searchResults.length == 0) {
-            // Returning a 404 error.
+            // Data is not found but still returning OK
             return new ResponseEntity<String>("No records found for the search condition.",
```
jdamerow: Let's return some proper xml here as well. Otherwise, it would throw clients off when they expect xml but just get text.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/ConceptSearch.java
```diff
@@ -126,4 +130,10 @@
 
         return new ResponseEntity<String>(msg.getXML(xmlEntries), HttpStatus.OK);
     }
+
+    @ExceptionHandler({ BindException.class })
+    public ResponseEntity<String> handleMethodArgumentTypeMismatch(BindException ex) {
```
jdamerow: where is that being thrown?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/ConceptSearchParameters.java
```diff
@@ -0,0 +1,129 @@
...
+    private String similar_to;
+    private String description;
+    @NotEmpty(message = "Word cannot be empty.")
+    private String word;
+    private String pos;
+    private String concept_list;
+    private String type_id;
+    private String creator;
+    private String modified_by;
+    private Integer number_of_records_per_page;
```
jdamerow: you probably want to validate a little bit more than just the word. what for example about pos value or operator?

mkarthik90:

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/ConceptSearchParameterValidator.java
```diff
@@ -0,0 +1,42 @@
+package edu.asu.conceptpower.servlet.rest;
+
+import org.springframework.stereotype.Component;
+import org.springframework.validation.Errors;
+import org.springframework.validation.Validator;
+
+import edu.asu.conceptpower.servlet.core.POS;
+
+@Component
+public class ConceptSearchParameterValidator implements Validator {
```
jdamerow: javadoc

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/ConceptSearchParameterValidator.java
```diff
@@ -0,0 +1,42 @@
...
+    @Override
+    public boolean supports(Class<?> clazz) {
+        return ConceptSearchParameters.class.isAssignableFrom(clazz);
+    }
+
+    @Override
+    public void validate(Object target, Errors errors) {
+        ConceptSearchParameters conceptSearchParameter = (ConceptSearchParameters) target;
+
+        if (conceptSearchParameter.getWord() == null || conceptSearchParameter.getWord().isEmpty()) {
```
jdamerow: you can use validationutils here (it becomes a one liner then)

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/ConceptSearchParameterValidator.java
```diff
@@ -0,0 +1,42 @@
...
+        } else if (!POS.posValues.contains(conceptSearchParameter.getPos())) {
+            errors.reject("pos", "Please enter correct pos value.");
+        }
+
+        if (conceptSearchParameter.getOperator() != null) {
+            if (SearchParamters.OP_OR.equalsIgnoreCase(conceptSearchParameter.getOperator())
+                    || SearchParamters.OP_AND.equalsIgnoreCase(conceptSearchParameter.getOperator())) {
+                errors.reject("operator", "Please enter correct operator value.");
+            }
+        }
```
jdamerow: I think we are not using localized messages for CP yet, but lets start doing it. I'll create a new ticket for that though. So for now, let's leave it the way it is.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/xml/XMLConstants.java
```diff
@@ -37,4 +38,5 @@
 	public final static String TYPE_URI_ATTR = "type_uri";
 	public final static String MATCHES = "matches";
     public final static String NUMBER_OF_RECORDS = "number_of_records";
+    public final static String ERROR_MESSAGE = "error_message";
 }
```
jdamerow: let's reformat this before merging

mkarthik90:

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/ConceptSearchParameterValidator.java
```diff
@@ -2,10 +2,19 @@
...
 import org.springframework.validation.Errors;
+import org.springframework.validation.ValidationUtils;
 import org.springframework.validation.Validator;
 
 import edu.asu.conceptpower.servlet.core.POS;
 
+/**
+ * This class validates the Concept power search parameters. Checks whether user
+ * has submitted word and pos details for searching the concept In addition, the
+ * validator checks for the operators (OR and AND)
```
jdamerow: - Conceptpower is one word. 
- Should be "Checks whether **a** user has submitted word and pos details **when** searching for a concept**.**" (period is missing at the end)
- periods is missing at the end of last sentence as well.

jdamerow:
Almost there.

mkarthik90:

jdamerow:
ok to test

jdamerow:
hm, well something is wrong:
https://diging-dev.asu.edu/conceptpower-review/rest/ConceptSearch?word=pony&pos=noun&operator=and
should give me a result.

Also, if you leave out operator it throws an error, which it shouldn't because it should use the default operator.

mkarthik90:
@jdamerow  Fixed review comments.

Operator by default is or
If operator is other than "and" or "or", the system throws error.

mkarthik90:
@jdamerow Merged with develop and addressed review comments

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/ConceptSearch.java
```diff
@@ -123,10 +127,15 @@ private void initBinder(WebDataBinder binder) {
...
                     page, numberOfRecordsPerPage);
-        } catch (LuceneException | IllegalAccessException ex) {
+        } catch (LuceneException ex) {
+            logger.error("Lucene Exception", ex);
             return new ResponseEntity<String>(ex.getMessage(), HttpStatus.INTERNAL_SERVER_ERROR);
-        } catch (IndexerRunningException e) {
-            return new ResponseEntity<String>(e.getMessage(), HttpStatus.SERVICE_UNAVAILABLE);
+        } catch (IllegalAccessException iae) {
+            logger.error("Illegal access exception", iae);
+            return new ResponseEntity<String>(iae.getMessage(), HttpStatus.BAD_REQUEST);
```
jdamerow: when is this exception thrown exactly?

jdamerow:
Where does it say what page is returned? This only seems to say how many results in total:
```<digitalHPS:pagination number_of_records="100" ></digitalHPS:pagination>```

File: Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/ConceptSearch.java (hunk 13 above)
Comment (mkarthik90) [inline Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/ConceptSearch.java]: That has not been thrown. I added it without noticing that. Removed that catch statements.

**Comment by mkarthik90** [comment]

File: Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/rest/ConceptSearch.java [hunk 14]
```diff
@@ -59,37 +75,73 @@
...
+     * @throws IllegalAccessException
+     * @throws IllegalArgumentException
      */
     @RequestMapping(value = "rest/ConceptSearch", method = RequestMethod.GET, produces = "application/xml; charset=utf-8")
-    public @ResponseBody ResponseEntity<String> searchConcept(HttpServletRequest req) {
-        Map<String, String[]> queryParams = req.getParameterMap();
+    public @ResponseBody ResponseEntity<String> searchConcept(
+            @Validated ConceptSearchParameters conceptSearchParameters,
+            BindingResult result)
+            throws IllegalArgumentException, IllegalAccessException {
```
jdamerow: Why are those thrown here? Are they caught anywhere?

mkarthik90: In this method we are using reflection technique to fetch the searchFields value. So it throws these exceptions.

jdamerow: Then you need to handle that in your controller and return a proper error message. I don't think an exception thrown here would return proper xml or json and error code, would it?

mkarthik90: yes I will do that

mkarthik90:

Added global exception handlers

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/servlet/util/ExceptionHandler.java
```diff
@@ -36,5 +36,17 @@ public String handle(Exception e) {
 		logger.error("ExceptionHandler caught exception.", e);
 		return "exception";
 	}
+	
+    @org.springframework.web.bind.annotation.ExceptionHandler(value = IllegalArgumentException.class)
+    public ResponseEntity<String> handleIllegalArgumentException(IllegalArgumentException ex) {
```
jdamerow: Please find me when you're back from lunch to about it.

mkarthik90:

This change has the following changes. 
1) Changes for package structure for implementing different
exception handlers. 
2) Added the code changes of CCP-171 into CCP-176. 
3) Review comment changes of CCP-171.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/util/ExceptionHandler.java
```diff
@@ -0,0 +1,28 @@
...
+ * 
+ * @author Julia Damerow
+ *
+ */
+@ControllerAdvice
+public class ExceptionHandler {
+	
+    private static final Logger logger = LoggerFactory.getLogger(ExceptionHandler.class);
+
+    @org.springframework.web.bind.annotation.ExceptionHandler(value = { Exception.class, IllegalArgumentException.class,
```
jdamerow: I"m pretty sure this has to go into the web package?

jdamerow:
There is a problem with the messages:
https://diging-dev.asu.edu/conceptpower-review/home/conceptsearch?word=John+Christopher+Willis+&pos=noun

jdamerow:
The search on the edit concept page doesn't return anything.
image

jdamerow:
And the search for creating wordnet wrappers doesn't work anymore either.

jdamerow:
On the edit concept page: hitting the enter key in the search for wordnet concepts doesn't submit the search (it just does nothing).

mkarthik90:
@jdamerow  Can you deploy this change in develop. I have added some logger debug statements to find the issue. I doubt if the numberOfLuceneResults from property file is not getting loaded.

jdamerow:
@mkarthik90 should be deployed now

mkarthik90:

Changes made only for concept list. If we make changes to all the string fields while indexing, it causes problems in other fields. For example if we index a wordnet id field as string field and convert it to lower case, we face problem while we try to fetch the data from wordnet manager, because wordnet manager checks whether wordnet id is of the form "WID". In our case it will be "wid". So to avoid confusion I just made change to concept list.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/ConceptManager.java
```diff
@@ -102,15 +101,31 @@ public ConceptEntry getWordnetConceptEntry(String wordnetId) throws LuceneExcept
      * java.lang.String, java.lang.String)
      */
     @Override
-    public ConceptEntry[] getConceptListEntriesForWord(String word, String pos, String conceptList)
+    public ConceptEntry[] getConceptListEntriesForWordPOSConceptList(String word, String pos, String conceptList)
```
jdamerow: given that the method is already called get**ConceptListEntries**ForWord... it seems doppeltgemoppelt (doublemoppeled) to attach ConceptList at the end of the method.

jdamerow:
The description needs to be shown for wordnet concepts in the edit concept screen. Otherwise it's hard to find out if the concept you're looking at is actually the one you need.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/ConceptManager.java
mkarthik90: @jdamerow Can I change it as getConceptListEntriesByWordAndPosAndConceptList?

mkarthik90:

jdamerow:
The messages are still broken:
https://diging-dev.asu.edu/conceptpower-review/home/conceptsearch?word=John+Christopher+Willis+&pos=noun

jdamerow:
The form should use the whole width of the page:
image

mkarthik90:

1) Fixed alignment issue
2) Fixed error message issue

jdamerow:
The labels don't need that much space:
image

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/core/ConceptEntry.java
```diff
@@ -10,12 +10,12 @@
 import javax.persistence.Entity;
 import javax.persistence.Id;
 
+import edu.asu.conceptpower.app.reflect.LuceneField;
+import edu.asu.conceptpower.app.reflect.SearchField;
+import edu.asu.conceptpower.rest.LuceneFieldNames;
+import edu.asu.conceptpower.rest.SearchFieldNames;
```
jdamerow: Those should not be in the rest package.

mkarthik90:

1) Changes for request mapping of rest interface.
2) Adjusted label and text fields.

jdamerow:
There are some issues here with the search. I'll merge it for now, because of the package structure change, but we need to fix those issues before we can deploy to production. Also, please test the whole application for more issues.

PR #142, opened 2016-11-04
[CCP-163-2] Fix for displaying alternative ids within concept

mkarthik90:


jdamerow:
Don't forget that your git comments should include the jira ticket number, not your branch name.

jdamerow:
ok to test

jdamerow:
I just notice, the search does not return alternative ids.

jdamerow:
I'll merge this so that Suraj can keep on working, but please add the alternative ids to the search (and any other rest api endpoint returning concepts) as well.

PR #146, opened 2016-11-17
[CCP-193] Rest call search fix

mkarthik90:
There were two issues in rest ful call concept search
1) The type_id has been included in the query even when
type_id is null. When type_id is null, the old code used a
String.valueOf so "null" string was included in query string.
This is now removed checking null for types field like all other
fields.

2) The default search operator was OR for concept search rest
call, where as the search for UI is AND. Made sure both the search
use AND for default operator. If operator is supplied by user, then
that operator will be overridden.

3) There was one more issue with operator. The operators supplied
to lucene query string should always be in upper case. So made changes
accordingly.

### DIFF
--- a/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/rest/SearchParamters.java
+++ b/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/rest/SearchParamters.java
@@ -2,7 +2,7 @@ package edu.asu.conceptpower.rest;
 
 public interface SearchParamters {
 
-	public final static String OPERATOR = "operator";
+    public final static String OPERATOR = "operator";
     public final static String OP_AND = "AND";
     public final static String OP_OR = "OR";
     public final static String PAGE = "page";
--- a/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/rest/SearchParamters.java
+++ b/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/rest/SearchParamters.java
@@ -3,8 +3,8 @@ package edu.asu.conceptpower.rest;
 public interface SearchParamters {
 
 	public final static String OPERATOR = "operator";
-	public final static String OP_AND = "and";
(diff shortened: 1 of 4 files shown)


Conceptpower+Spring/src/main/java/edu/asu/conceptpower/rest/SearchParamters.java
```diff
@@ -3,8 +3,8 @@
 public interface SearchParamters {
 
 	public final static String OPERATOR = "operator";
-	public final static String OP_AND = "and";
-	public final static String OP_OR = "or";
+    public final static String OP_AND = "AND";
```
jdamerow: then reformat the file please

jdamerow:
ok to test

jdamerow:
Two things: 

1. the result order is different from the order in the search interface, which is odd since they should do the same thing:
https://diging-dev.asu.edu/conceptpower-review/rest/ConceptSearch?word=pony&pos=noun

2. alternativeIds are missing

mkarthik90:
Currently we have alternative ids in Concept ID lookup. I will including alternative ids in all the rest calls as  a part of CCP-185

jdamerow:
Ok, so there is an architectural issue here that I missed before regarding alternative ids. Of course, alternative ids always have to be included for every returned concept entry. The following changes are required:

1. When indexing, the alternative id field has to be filled with the CP id and wordnet ids. General wordnet ids are a special case and are left our here.
2. For concepts not retrieved through Lucene, the concept manager has to fill the alternative ids filled when retrieving concepts. 
3. The concept manager also has to add general wordnet ids if it applies. Also the logic for general wordnet ids has to be done in the concept manager.
3. The controllers shouldn't have to worry about those things at all and just marshall objects.

Sorry, don't know how I missed that.

mkarthik90:

The alternative id changes will be completed as part of CCP-185

PR #148, opened 2016-12-21
Story/CCP-186

mkarthik90:


mkarthik90:
@jdamerow Please review the changes.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptMergeController.java
```diff
@@ -0,0 +1,53 @@
...
+@Controller
+public class ConceptMergeController {
+
+    @Autowired
+    private IConceptManager conceptManager;
+
+    @Autowired
+    private IConceptMergeService conceptMergeService;
+
+    @RequestMapping(value = "prepareMergeConcept")
```
jdamerow: Why would that not be in the authorization requiring part of Conceptpower?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/bean/ConceptsMergeBean.java
```diff
@@ -0,0 +1,164 @@
...
+    private Set<String> conceptList;
+    private Set<String> description;
+    private Set<String> synonymsids;
+    private Set<String> conceptTypeIdList;
+    private Set<String> equals;
+    private Set<String> similar;
+    private String selectedTypeId;
+    private String selectedTypeName;
+    private ConceptType[] types;
+    private String selectedConceptId;
```
jdamerow: all these variables should not be called "word" or "conceptList" if their type is ```Set```. This is very confusing.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
```diff
@@ -0,0 +1,297 @@
...
+    private ConceptTypesService conceptTypesService;
+
+    @Override
+    public ConceptsMergeBean prepareMergeConcepts(List<ConceptEntry> conceptEntries,
+            ConceptsMergeBean conceptsMergeBean) {
+        for (ConceptEntry entry : conceptEntries) {
+
+            if (!isNullOrEmpty(entry.getPos())) {
+                conceptsMergeBean.setSelectedPosValue(entry.getPos().toLowerCase());
+            }
```
jdamerow: Instead of all this ```isNullOrEmpty``` method callsed, why don't you just check for null/empty when adding to the new set?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/bean/ConceptsMergeBean.java
```diff
@@ -0,0 +1,164 @@
...
+
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import edu.asu.conceptpower.app.core.POS;
+import edu.asu.conceptpower.core.ConceptType;
+
+public class ConceptsMergeBean {
```
jdamerow: I would not combine the possible values for the new merged concept with the merged concept. Just use an existing concept bean class or create a separate one. It'll make this all much more understandable.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
```diff
@@ -0,0 +1,297 @@
...
+            equalsBuffer.append(equals);
+        }
+        entry.setEqualTo(equalsBuffer.toString().trim());
+
+        StringBuffer similarBuffer = new StringBuffer();
+        prefix = "";
+        for (String similarTo : conceptMergeBean.getSimilar()) {
+            similarBuffer.append(prefix);
+            prefix = ",";
+            similarBuffer.append(similarTo);
```
jdamerow: instead of all these for loops just use a stringjoiner or even better the collector as shown here:
https://docs.oracle.com/javase/8/docs/api/java/util/StringJoiner.html

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptMergeController.java
mkarthik90: @jdamerow  This method will be called once user selects the concepts to be merged (by checking the checbox). This method is used for obtaining all the concepts selected for merge by the user and creating a new concept entry (merged concept entry). I am not sure how to add this entire logic to authorization part of CCP.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/bean/ConceptsMergeBean.java
mkarthik90: @jdamerow  I created this new class inorder to easily bind the values to the front end. Once user clicks on prepare merge concepts a UI screen appears with the merged concept details. So I thought we could have a new bean that maps to this screen and user can edit the values of the merged concept. Should I rename this new bean class file ?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/bean/ConceptsMergeBean.java
mkarthik90: Renamed

mkarthik90:
@jdamerow  Addressed review comments. Addressed comment from hipchat as well.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptMergeController.java
jdamerow: url should be ```auth/...``` like all the others.

Also, instead of camel case, I'd use a path. E.g. ```auth/concepts/merge``` and ```auth/concepts/merge/prepare``` (if that doesn't overlap with any other controllers).

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/bean/ConceptsMergeBean.java
jdamerow: But these are two different things, and you can just as easily use one object for the input values for the page and another one for the submitted values. Or do you need all type ids and all list names in the submitted object to create the new merged concept?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
```diff
@@ -0,0 +1,289 @@
...
+        Set<String> conceptLists = new HashSet<>();
+        for (ConceptList conceptList : allLists) {
+            conceptLists.add(conceptList.getConceptListName());
+        }
+        return conceptLists;
+    }
+
+    private Set createSet(String value, Set collection) {
+
+        if (isNullOrEmpty(value)) {
```
jdamerow: you really don't need its own method for that.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
```diff
@@ -0,0 +1,289 @@
...
+        }
+
+        if (collection == null) {
+            collection = new HashSet();
+        }
+        collection.add(value);
+        return collection;
+    }
+
+    private ConceptType[] getTypesByTypeId(Set<String> conceptTypesId) {
```
jdamerow: use plural for lists please (variables, methods, etc).

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
```diff
@@ -0,0 +1,289 @@
...
+
+    private Set<String> combineSynonymIds(Set<String> existingSynonymId, String newSynonymIds) {
+
+        if (newSynonymIds == null) {
+            return existingSynonymId;
+        }
+
+        if (newSynonymIds.trim().isEmpty()) {
+            return existingSynonymId;
+        }
```
jdamerow: this can be combined with the previous if statement

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
```diff
@@ -0,0 +1,289 @@
...
+
+        String[] synonymIds = newSynonymIds.split(",");
+
+        for (String synonymId : synonymIds) {
+            if (existingSynonymId == null) {
+                existingSynonymId = new HashSet<>();
+            }
```
jdamerow: no reason to check this every time. one check at the beginning is enough. but also, this can all be a one liner:

```new HashSet<T>(Arrays.asList(array));```

File: Conceptpower+Spring/src/main/webapp/WEB-INF/views/home.jsp [hunk 10]
```diff
@@ -111,12 +160,17 @@ function showFormProcessing() {
   action="${pageContext.servletContext.contextPath}/home/conceptsearch"
   method='get' commandName='conceptSearchBean'>
   <form:errors path="luceneError"></form:errors>
+  
+    <div id="mergeError" class="alert alert-danger">
+      Please select atleast two concepts to merge.
```
jdamerow: "at least" are two words.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptMergeController.java
mkarthik90: Got it

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/bean/ConceptsMergeBean.java
mkarthik90: I tried to use two different objects, but the JSTL spring tags don't support two different objects. One for input and another one for submitted values. For example, the <form:input> tag has only one attribute path with which the value will be bound to the server. Two objects scenario may hold good only for select boxes, where the set of objects can be iterated from a different object but the one selected value can be loaded into the submitted object. But we have POS, Concept type, concept list as a drop down, rest of the values are input or text area. Not sure how to use two different objects here (one for input and another for submitted output). In addition, I checked the ConceptEditBean, we have used the same object for both input and submitted values. Please let me know if there is a way to handle this in JSTL

mkarthik90:

mkarthik90:
@jdamerow  Addressed review comments. Input values are loaded to model and output to ConceptsMergeBean object.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/constants/SearchFieldNames.java
```diff
@@ -17,4 +17,5 @@
 	public final static String TYPE_URI = "type_uri";
```
jdamerow: formatting

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/POS.java
```diff
@@ -2,7 +2,9 @@
```
jdamerow: formatting

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/POS.java
```diff
@@ -13,4 +15,13 @@
 	public final static String OTHER = "other";
 	
 	public final static List<String> posValues = new ArrayList<String>(Arrays.asList(new String[] {POS.NOUN, POS.VERB, POS.ADVERB, POS.ADJECTIVE, POS.OTHER}));
+
+    public static Map<String, String> getPosMap() {
```
jdamerow: this should not be in an interface but in a message file.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
```diff
@@ -0,0 +1,254 @@
...
+            // Add
+            ConceptEntry entry = new ConceptEntry();
+            getConceptEntryFromConceptMergeBean(entry, conceptsMergeBean);
+            conceptManager.addConceptListEntry(entry, userName);
+        } else {
+            // Update
+            ConceptEntry entry = conceptManager.getConceptEntry(conceptsMergeBean.getSelectedConceptId());
+            getConceptEntryFromConceptMergeBean(entry, conceptsMergeBean);
+            conceptManager.storeModifiedConcept(entry, userName);
+        }
```
jdamerow: I'd first create a new concept and then delete the old ones. Otherwise, if anything goes wrong (e.g. server crashes right after deletion before creation of new concepts) the old concepts are gone but there are is no new merged concept yet.

jdamerow: Ideally we would use a transaction, but I think it's overkill right now.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    private String createConceptWrapperById(String conceptId, String userName, String conceptListName)
+            throws IllegalAccessException, DictionaryDoesNotExistException, DictionaryModifyException, LuceneException,
+            IndexerRunningException {
+        ConceptEntry entry = conceptManager.getConceptEntry(conceptId);
+        entry.setConceptList(conceptListName);
+        // Not adding other details of merged concepts. Since its not relevant
+        // to add these details here.
+        return conceptManager.addConceptListEntry(entry, userName);
```
jdamerow: you should still create a proper concept wrapper. We'll probably have restore, search, etc. functions for deleted concepts later on and for that we would need complete concepts.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
```diff
@@ -0,0 +1,254 @@
...
+
+        String wordnetIds = conceptMergeBean.getWordnetIds().stream().map(i -> i.toString())
+                .collect(Collectors.joining(prefix));
+        entry.setWordnetId(wordnetIds);
+
+        addAlternativeIdsToEntry(entry, conceptMergeBean);
+        addMergedIdsToEntry(entry, conceptMergeBean);
+    }
+
+    private void addWordNetIdsToEntry(ConceptEntry entry, ConceptsMergeBean conceptMergeBean) {
```
jdamerow: is this method still used?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
```diff
@@ -0,0 +1,254 @@
...
+                    .getConceptTypeByConceptId(selectedId)) {
+                alternativeIds.add(conceptId);
+                ConceptEntry conceptEntry = conceptManager.getConceptEntry(conceptId);
+                alternativeIds.addAll(conceptEntry.getAlternativeIds());
+            }
+        }
+        entry.setAlternativeIds(alternativeIds);
+    }
+
+    private void addMergedIdsToEntry(ConceptEntry entry, ConceptsMergeBean conceptMergeBean) {
```
jdamerow: you can replace this method probably with a java 8 stream -> filter -> map -> collect as well (just add a filter).

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
```diff
@@ -0,0 +1,254 @@
...
+        return conceptLists;
+    }
+
+    private Set createSet(String value, Set collection) {
+        if (value == null) {
+            return null;
+        }
+        if (value.trim().isEmpty()) {
+            return null;
+        }
```
jdamerow: these two ifs at least can be in one if

Conceptpower+Spring/src/main/webapp/WEB-INF/views/auth/concepts/MergeConcepts.jsp
```diff
@@ -0,0 +1,92 @@
...
+<title>Merge Concept</title>
+<form:form action="${pageContext.servletContext.contextPath}/auth/concepts/merge" method='post' modelAttribute="conceptsMergeBean">
+
+    <h1>Merge concepts</h1>
+
+    <table>
+        <tr>
+            <td class="col-sm-1"> Concept Id </td>
+        </tr>
+        <c:forEach var="conceptId" items="${conceptsMergeBean.conceptIds}">
```
jdamerow: Isn't there an input tag radiobuttons (that prints all of them)? Also, separating the radio button and its label in two different tds seems (without having tested it yet) like it might look a little funny on big screens (tds being pulled apart).

mkarthik90:

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/POSMessage.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.ResourceBundle;
+
+import javax.annotation.PostConstruct;
+
+import org.springframework.context.annotation.PropertySource;
+import org.springframework.stereotype.Component;
+
+@Component
+@PropertySource("classpath:pos.properties")
+public class POSMessage {
```
jdamerow: this seems overkill. you should be able to just use the messages right in the jsp file, similar to this:
http://stackoverflow.com/questions/11792734/accessing-dynamic-resource-message

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
```diff
@@ -0,0 +1,235 @@
...
+                alternativeIds.add(conceptId);
+                ConceptEntry conceptEntry = conceptManager.getConceptEntry(conceptId);
+                alternativeIds.addAll(conceptEntry.getAlternativeIds());
+            }
+        }
+        entry.setAlternativeIds(alternativeIds);
+    }
+
+    private Predicate<String> isSpecificWordnetConcept(String selectedId) {
+        return p -> !p.equalsIgnoreCase(selectedId)
```
jdamerow: I would use a more descriptive variable name here. Since ```p``` will always refer to an id, and ```id``` is not much longer than ```p```, you should use that here. It is not clear, why you would call the variable ```p``` here but you code would be much more understandable with a proper variable name.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/POSMessage.java
mkarthik90: @jdamerow I went through the docs of spring:message. We don't have any method for fetching all the keys. Using spring:message we will be able to fetch the values from resource file based on the keys. So should I retrive the keys from Java end by using keySet() in resourcebundle. And iterate those keys in jsp and fetch the data from properties file? Is it advisable?

mkarthik90:
@jdamerow Fixed review comments. I have replied to one of your comment for POS dropdown.

jdamerow:
please resolve conflicts

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/POSMessage.java
jdamerow: You can get all POS from the interface that already exists and then get for each POS the corresponding message. You don't need an extra properties/messages file for POS values.

jdamerow: Does this make sense?

mkarthik90:
@jdamerow  Resolved merge conflicts and addressed review comments.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
```diff
@@ -0,0 +1,235 @@
...
+            throws IllegalAccessException, DictionaryDoesNotExistException, DictionaryModifyException, LuceneException,
+            IndexerRunningException {
+        ConceptEntry entry = conceptManager.getConceptEntry(conceptId);
+        // Creating concept wrapper with all the values, because in future we
+        // will be including manipulations on deleted wrappers as well.
+        getConceptEntryFromConceptMergeBean(entry, conceptsMergeBean);
+        return conceptManager.addConceptListEntry(entry, userName);
+    }
+
+    public void getConceptEntryFromConceptMergeBean(ConceptEntry entry, ConceptsMergeBean conceptMergeBean) {
```
jdamerow: should not be called ```get...``` since it doesn't "get" anything (no return value).

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
```diff
@@ -0,0 +1,235 @@
...
+        addMergedIdsToEntry(entry, conceptMergeBean);
+    }
+
+    private void addAlternativeIdsToEntry(ConceptEntry entry, ConceptsMergeBean conceptMergeBean) {
+        List<String> conceptIds = conceptMergeBean.getConceptIds();
+        String selectedId = conceptMergeBean.getSelectedConceptId();
+        Set<String> alternativeIds = new HashSet<>();
+
+        for (String conceptId : conceptIds) {
+            if (!conceptId.equalsIgnoreCase(selectedId) || ConceptTypes.SPECIFIC_WORDNET_CONCEPT == conceptTypesService
```
jdamerow: i'm confused. why do you exclude those two cases?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
```diff
@@ -0,0 +1,235 @@
...
+
+    private Predicate<String> isSpecificWordnetConcept(String selectedId) {
+        return id -> !id.equalsIgnoreCase(selectedId)
+                || ConceptTypes.SPECIFIC_WORDNET_CONCEPT == conceptTypesService.getConceptTypeByConceptId(selectedId);
+    }
+
+    private void addMergedIdsToEntry(ConceptEntry entry, ConceptsMergeBean conceptMergeBean) {
+        String prefix = ",";
+        String selectedId = conceptMergeBean.getSelectedConceptId();
+        String mergedIds = conceptMergeBean.getConceptIds().stream().map(i -> i.toString())
```
jdamerow: here too. why do you filter those?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
```diff
@@ -0,0 +1,235 @@
...
+
+    @Override
+    public Set<String> getAllConceptLists() {
+        List<ConceptList> allLists = conceptListManager.getAllConceptLists();
+        Set<String> conceptLists = new HashSet<>();
+        for (ConceptList conceptList : allLists) {
+            conceptLists.add(conceptList.getConceptListName());
+        }
+        return conceptLists;
+    }
```
jdamerow: why has this its own method? If you can use the list of ConceptLists itself, it should be a one liner in java 8 to turn that into a collection of strings.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptMergeController.java
```diff
@@ -0,0 +1,64 @@
...
+    @Autowired
+    private IConceptManager conceptManager;
+
+    @Autowired
+    private IConceptMergeService conceptMergeService;
+
+    @Autowired
+    private IConceptTypeManger conceptTypesManager;
+
+    @RequestMapping(value = "auth/concepts/merge/prepare")
```
jdamerow: make this just auth/concepts/merge

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptMergeController.java
```diff
@@ -0,0 +1,64 @@
...
+        mav.setViewName("/auth/conceptMerge");
+        return mav;
+    }
+
+    @RequestMapping(value = "auth/concepts/merge", method = RequestMethod.POST)
+    public String mergeConcept(ModelMap model, @ModelAttribute("conceptsMergeBean") ConceptsMergeBean conceptsMergeBean,
+            BindingResult result, Principal principal) throws IllegalAccessException, LuceneException,
+            IndexerRunningException, DictionaryDoesNotExistException, DictionaryModifyException {
+        conceptMergeService.mergeConcepts(conceptsMergeBean, principal.getName());
+        return "redirect:/login";
```
jdamerow: why do you redirect to login?

Conceptpower+Spring/src/main/webapp/WEB-INF/views/auth/concepts/MergeConcepts.jsp
```diff
@@ -0,0 +1,95 @@
...
+    <br />
+    <br />
+
+    <table class="width: 100%;">
+        <tr>
+            <td><input type="submit" name="edit" id="edit"
+                value="Store merged concept" class="btn btn-primary"></td>
+
+            <td><a
+                href="${pageContext.servletContext.contextPath}/auth/concepts/canceledit?fromHomeScreen=true"><input
```
jdamerow: why don't you just put in the URL of the homescreen directly?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptMergeController.java
mkarthik90: We have one more request mapping auth/concepts/merge. Since this step just prepares the conceptmergebean object by obtaining the values from selected concept entries I had the requestmapping as auth/concepts/merge/prepare. When user edits all the values for conceptmergebean and click on merge auth/concepts/merge will be called.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptMergeController.java
mkarthik90: Once concepts are merge, user will be navigated to welcome screen so thats why I was redirecting to /login. And redirect was used because even if the client tries to refresh the page only the welcome page will be loaded and merge operation will not be triggered again.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptMergeController.java
jdamerow: but isn't the other one a post request?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptMergeController.java
jdamerow: but shouldn't it redirect to the search (not login)?

mkarthik90: ok i will do that

jdamerow: or ```/```

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptMergeController.java
mkarthik90: Got it. I will change the mapping

mkarthik90:
@jdamerow  Addressed review comments and added validator

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
```diff
@@ -0,0 +1,170 @@
...
+            throws IllegalAccessException, DictionaryDoesNotExistException, DictionaryModifyException, LuceneException,
+            IndexerRunningException {
+        ConceptEntry entry = conceptManager.getConceptEntry(conceptId);
+        // Creating concept wrapper with all the values, because in future we
+        // will be including manipulations on deleted wrappers as well.
+        createConceptEntryFromConceptsMergeBean(entry, conceptsMergeBean);
+        return conceptManager.addConceptListEntry(entry, userName);
+    }
+
+    public void createConceptEntryFromConceptsMergeBean(ConceptEntry entry, ConceptsMergeBean conceptMergeBean) {
```
jdamerow: this is also not creating a new object (you're passing it as parameter). maybe 'fill' or 'transfer content' (then without the 'to concept entry to ...' to keep it short.

jdamerow:
ok to test

jdamerow:
you forgot to change the compiler version maven uses, hence the build fails.

mkarthik90:

jdamerow:
Prepare merge button should be by the results (maybe smaller and between "Results" heading and table), and not right next to the search box (especially when the user hasn't searched yet, this looks odd as the search bar doesn't go over the whole width of the page).
image

jdamerow:
image
**Concept** should only have one word in there (maybe just use the longest?). also, why does it replace the blank with an underscore?
**POS** should be default be the POS of the first concept. If there are different POSs there should be a warning that the user is about to merge concepts that have different POSs.
**Description** should not start with a comma. And rather than appended by comma, each description should start a new paragraph.
**Concept Type**, if there is a concept among the concepts that has one that should be there. If there are several, use the first and show a warning.

jdamerow:
The page should also show how many concepts are being merged and which ones (id and concept name).

jdamerow:
Merge failed:
Sorry but something went horribly wrong...!

mkarthik90:

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
```diff
@@ -32,27 +32,62 @@
...
+        if (posWarning) {
+            errorBuilder.append("<p>There are difference in pos within the merging concepts.</p>");
+        }
+
+        if (typeWarning) {
+            errorBuilder.append("<p>Merging concepts have different concept type.</p>");
+        }
+
+        conceptsMergeBean.setErrorMessages(errorBuilder.toString());
+
```
jdamerow: don't put html in your beans (especially in the service class), it'll make it not reusable for situations where you don't want html (e.g. when using through a rest api).

jdamerow: Also, use an error classes instead of just one string (with field code an error code). Then you can display errors on the page nicely formatted and depending on output.

mkarthik90:

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
```diff
@@ -0,0 +1,224 @@
...
+            } else if (entry.getTypeId() != null) {
+                typeWarning = true;
+            }
+
+            if (entry.getWordnetId() != null && !entry.getWordnetId().isEmpty()) {
+                String[] wordNetIds = entry.getWordnetId().split(",");
+                conceptsMergeBean.getMergedIds()
+                        .addAll(Arrays.asList(wordNetIds).stream()
+                                .filter(wordNetId -> !wordNetId.trim().equalsIgnoreCase("")).map(wordNetId -> wordNetId)
+                                .collect(Collectors.toSet()));
```
jdamerow: Why are wordnet ids added to the mergedId list here, but the actually merged concepts are added later? It doesn't make sense and takes two things apart that belong together. Is there a reason?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
```diff
@@ -0,0 +1,224 @@
...
+            }
+            if (entry.getSimilarTo() != null && !entry.getSimilarTo().isEmpty()) {
+                conceptsMergeBean.getSimilarValues().add(entry.getSimilarTo());
+            }
+
+            if (entry.getTypeId() != null && conceptsMergeBean.getSelectedTypeId() == null) {
+                conceptsMergeBean.setSelectedTypeId(entry.getTypeId());
+            } else if (entry.getTypeId() != null) {
+                typeWarning = true;
+            }
```
jdamerow: wouldn't that also result in a warning if two concepts have the same type?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
```diff
@@ -0,0 +1,224 @@
...
+            CPError errroMessage = new CPError("error_message_101",
+                    "There are difference in pos within the merging concepts.");
+            conceptsMergeBean.getErrorMessages().add(errroMessage);
+        }
+
+        if (typeWarning) {
+            CPError errroMessage = new CPError("error_message_102",
+                    "Merging concepts have different concept type.");
+            conceptsMergeBean.getErrorMessages().add(errroMessage);
+        }
```
jdamerow: The warnings should show the different POSs/types used. I would get rid of the boolean flag, and add the warnings right where they are noticed (adding the POSs/types).

Conceptpower+Spring/src/main/webapp/WEB-INF/views/home.jsp
```diff
@@ -150,17 +204,26 @@ function showFormProcessing() {
...
-
+<c:if test="${not empty conceptSearchBean.foundConcepts}">
 <h3>Results</h3>
 
-<c:if test="${not empty conceptSearchBean.foundConcepts}">
+    <sec:authorize access="isAuthenticated()">
+        <div class="row">
+            <div class="col-sm-2">
+                <button id="prepareMergeConcept" class="btn-sm btn-action">Prepare Merge Concepts</button>
+                <button id="mergeConcept" class="btn-sm btn-action" onclick="mergeConcepts();">Merge Concepts</button>
```
jdamerow: If the buttons still stick to the below table as they did when i saw it on your computer, make sure to put some padding there.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
mkarthik90: I was thinking if that would add up multiple duplicate error messages. Else I need to have a boolean variable to mark that error has been already added.

mkarthik90:

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
```diff
@@ -0,0 +1,211 @@
...
+            ConceptEntry entry = new ConceptEntry();
+            fillConceptEntry(entry, conceptsMergeBean);
+            conceptManager.addConceptListEntry(entry, userName);
+        } else {
+            // Update
+            ConceptEntry entry = conceptManager.getConceptEntry(conceptsMergeBean.getSelectedConceptId());
+            fillConceptEntry(entry, conceptsMergeBean);
+            // Wordnet id has been added to alternative ids field. So setting it
+            // to blank
+            entry.setWordnetId("");
```
jdamerow: just don't do this and we should be fine.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptMergeController.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    @RequestMapping(value = "auth/concepts/merge", method = RequestMethod.POST)
+    public ModelAndView mergeConcept(
+            @ModelAttribute("conceptsMergeBean") @Validated ConceptsMergeBean conceptsMergeBean, BindingResult result,
+            Principal principal) throws IllegalAccessException, LuceneException, IndexerRunningException,
+                    DictionaryDoesNotExistException, DictionaryModifyException {
+        ModelAndView mav = new ModelAndView();
+        if (result.hasErrors()) {
+            mav.setViewName("forward:/auth/conceptMerge");
```
jdamerow: is there a reason you use 'forward' here and not 'redirect'?

jdamerow: also, why doesn't that show the merge screen again?

jdamerow: isn't that a tiles view name rather than a url?

mkarthik90: Yes this did not work. I set the view to the merge screen and loaded the drop down values alone. I did not redirect to the get method, because prepareMerge does some processing before directing to the view component. So I just set the value on tiles and returned the model.

jdamerow:
Apparently for some concepts when they get retrieved from the database (maybe before there were alternative ids), the alternative ids field is null and ```prepareMergeConcepts``` throws a null pointer.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
```diff
@@ -0,0 +1,207 @@
...
+        boolean posWarning = false;
+        conceptsMergeBean.clear();
+        for (ConceptEntry entry : conceptEntries) {
+            if (entry.getWord() != null && !entry.getWord().isEmpty()) {
+                conceptsMergeBean.setWord(conceptsMergeBean.getWord() == null
+                        || conceptsMergeBean.getWord().length() < entry.getWord().length() ? entry.getWord()
+                                : conceptsMergeBean.getWord());
+                
+                String conceptIdWithName = entry.getWord().trim() + " -- " + entry.getId();
+                conceptsMergeBean.getConceptNamesWithIds().add(conceptIdWithName);
```
jdamerow: is this purely for presentation purposes on the merge site, that should ideally not be done in the business logic. This has nothing to do with business logic. Don't you have all selected concepts in the merge object, and can use that directly in the jsp page?

jdamerow:
If there is a validation error in the merge bean, the merged concepts are not shown anymore on the merge page that shows the validation errors.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
mkarthik90: yes i do have. But the the details are split.  ConceptMergeBean contains the conceptIds and wordnetids seperately. Also the word details are not stored in the conceptMergeBean. I will create a map that has the key as ConceptId (or) wordnetid and value as the word. I will iterate the map in the UI and show what concepts are getting merged

mkarthik90:
@jdamerow  I am not able to find the validation errors issue. I tested by not providing the POS or type values. The error message shows up along with the merged concept details. I checked on the dev server. Let me know if you mean something different.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
jdamerow: You already have the concept list in the controller, why don't you just put it in the model to be used in the jsp page?

Conceptpower+Spring/src/main/webapp/WEB-INF/views/auth/concepts/MergeConcepts.jsp
```diff
@@ -0,0 +1,181 @@
+<%@ taglib uri="http://www.springframework.org/tags" prefix="spring"%>
+<%@ taglib prefix="sec"
+    uri="http://www.springframework.org/security/tags"%>
+<%@ taglib prefix="c" uri="http://java.sun.com/jsp/jstl/core"%>
+<%@ taglib prefix="form" uri="http://www.springframework.org/tags/form"%>
+<head>
```
jdamerow: is there a closing tag anywhere? why is that here?

Conceptpower+Spring/src/main/webapp/WEB-INF/views/auth/concepts/MergeConcepts.jsp
```diff
@@ -0,0 +1,181 @@
...
+<title>Merge Concept</title>
```
jdamerow: some here, why is that here?

Conceptpower+Spring/src/main/webapp/WEB-INF/views/auth/concepts/MergeConcepts.jsp
```diff
@@ -0,0 +1,181 @@
...
+                </p>
+            </c:forEach>
+        </div>
+    </c:if>
+    
+    <div class="alert alert-info alert-dismissable">
+        The following concepts are being merged.
+        <ul class="list-group">
+            <c:forEach var="conceptEntry" items="${conceptEntries}">
+                <li class="list-group-item"><c:out value="${conceptEntry.id}"/> -- <c:out value="${conceptEntry.word}"/></li>
```
jdamerow: Let's make that:
```[id] word``` and on hover show description

File: Conceptpower+Spring/src/main/webapp/WEB-INF/views/auth/concepts/MergeConcepts.jsp [hunk 41]
```diff
@@ -0,0 +1,181 @@
...
+        <c:when test="${not empty conceptsMergeBean.conceptIds }">
+            <label class="col-sm-1 control-label">Concept Id</label>
+            <div class="col-sm-11">
+                <!-- If none of the values are selected, we create a new concept id for the merged concept -->
+                <form:radiobuttons path="selectedConceptId" items="${conceptsMergeBean.conceptIds}" element="label class='radio-inline'" />
+            </div>
+        </c:when>
+        <c:otherwise>
+            <div class="col-sm-12">
+                New Id will be generated for merged concepts.
```
jdamerow: "A new id will be generated for the merged concepts."

Conceptpower+Spring/src/main/webapp/WEB-INF/views/home.jsp
```diff
@@ -150,17 +204,26 @@ function showFormProcessing() {
...
 <hr>
-
+<c:if test="${not empty conceptSearchBean.foundConcepts}">
 <h3>Results</h3>
 
-<c:if test="${not empty conceptSearchBean.foundConcepts}">
+    <sec:authorize access="isAuthenticated()">
+        <div class="row">
+            <div class="col-sm-2">
+                <button id="prepareMergeConcept" class="btn-sm btn-action" style="margin-bottom: 15px;">Prepare Merge Concepts</button>
```
jdamerow: Button should say: "Select concepts to merge."

Conceptpower+Spring/src/main/webapp/WEB-INF/views/home.jsp
```diff
@@ -150,17 +204,26 @@ function showFormProcessing() {
...
+                <button id="mergeConcept" class="btn-sm btn-action" style="margin-bottom: 15px;" onclick="mergeConcepts();">Merge Concepts</button>
```
jdamerow: Button should say: "Merge selected concepts."

Conceptpower+Spring/src/main/webapp/WEB-INF/views/home.jsp
```diff
@@ -150,17 +204,26 @@ function showFormProcessing() {
...
+    </sec:authorize>
+
   <table cellpadding="0" cellspacing="0"
     class="table table-striped table-bordered" id="conceptSearchResult">
     <thead>
       <tr>
         <sec:authorize access="isAuthenticated()">
           <th></th>
           <th></th>
+          <th>Merge</th>
```
jdamerow: Remove column header.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptsMergeBeanValidator.java
```diff
@@ -0,0 +1,26 @@
...
+    public boolean supports(Class<?> clazz) {
+        return ConceptsMergeBean.class.isAssignableFrom(clazz);
+    }
+
+    @Override
+    public void validate(Object target, Errors errors) {
+        ValidationUtils.rejectIfEmpty(errors, "word", "word", "Word cannot be empty.");
+        ValidationUtils.rejectIfEmpty(errors, "selectedPosValue", "selectedPosValue", "POS cannot be empty.");
+        ValidationUtils.rejectIfEmpty(errors, "selectedListName", "selectedListName", "Concept List cannot be empty.");
+        ValidationUtils.rejectIfEmpty(errors, "selectedTypeId", "selectedTypeId", "Concept Type cannot be empty.");
```
jdamerow: only whitespace is also not allowed

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/bean/ConceptsMergeBean.java
```diff
@@ -0,0 +1,164 @@
...
+    private String selectedPosValue;
+    private String selectedListName;
+    private String selectedTypeId;
+    private String selectedTypeName;
+    private String selectedConceptId;
+    private List<String> descriptions = new ArrayList<>();
+    private Set<String> synonymsids = new HashSet<>();
+    private Set<String> equalsValues = new HashSet<>();
+    private Set<String> similarValues = new HashSet<>();
+    private Set<String> mergedIds = new HashSet<>();
```
jdamerow: Isn't mergedIds and selectedConceptIds the same thing?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptsMergeBeanValidator.java
jdamerow: and also make sure that ```merged ids``` contains all ids of merged concepts (at least 2) and that

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/bean/ConceptsMergeBean.java
mkarthik90: MergedIds contains the details of what concepts are getting merged. 

SelectedConceptId contains the id to which the merged concept details needs to be stored. If the selectedConceptId is blank or null, we store the merged concept details in a new concept.

ConceptIds will contain only the CCP concept Id and not the wordnet concept id. This field is used for showing the user with the list of concepts to which the merged concept can be stored. So wordnet ids are removed from this list.

jdamerow:
as just discussed, remove wordnet ids and concept ids in merge bean and resolve conflicts

mkarthik90:
@jdamerow  Addressed review comment and resolved conflicts.

mkarthik90:
@jdamerow Addressed review comments. Please let me know if there are any other comments.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptsMergeBeanValidator.java
```diff
@@ -0,0 +1,34 @@
...
+        ValidationUtils.rejectIfEmptyOrWhitespace(errors, "word", "word", "Word cannot be empty.");
+        ValidationUtils.rejectIfEmptyOrWhitespace(errors, "selectedPosValue", "selectedPosValue",
+                "POS cannot be empty.");
+        ValidationUtils.rejectIfEmptyOrWhitespace(errors, "selectedListName", "selectedListName",
+                "Concept List cannot be empty.");
+        ValidationUtils.rejectIfEmptyOrWhitespace(errors, "selectedTypeId", "selectedTypeId",
+                "Concept Type cannot be empty.");
+
+        ConceptsMergeBean conceptMergeBean = (ConceptsMergeBean) target;
+        if (conceptMergeBean.getMergeIds() != null && conceptMergeBean.getMergeIds().size() < 2) {
```
jdamerow: what if ```getMergeIds``` is null?

mkarthik90: @jdamerow  MergedIds cannot be null. This is because mergedids are instantiated in ConceptMergeBean class. Do you still want me to check for the null condition? I am not sure if I am missing something here.

mkarthik90:
@jdamerow Validated merged id field for null value as well.

jdamerow:
ok to test

jdamerow:
The hover on the concepts to be merged is still missing (showing description on hover).

jdamerow:
And while you're at it, would you please remove the periods from the "Select concepts for merge" and "Merge selected concepts" buttons (I was either too confused or too grammatically correct (ending my sentence with a period...)). ;)

jdamerow:
So, I merged three "Einstein" concepts and now it doesn't find Einstein anymore...

mkarthik90:

jdamerow:
retest this please

jdamerow:
hm, that does not seem to have worked:
image

jdamerow:
Also, for some reason. If you take the URL for merging concepts, e.g. https://diging-dev.asu.edu/conceptpower-review/auth/concepts/merge?mergeIds=WID-10841405-N-01-Beethoven,%20WID-10841405-N-02-van_Beethoven,%20WID-10841405-N-03-Ludwig_van_Beethoven
and go to that URL when you're not logged in, it says something went horribly wrong after logging in.

Steps to reproduce:
1. in incognito window go to https://diging-dev.asu.edu/conceptpower-review/auth/concepts/merge?mergeIds=WID-10841405-N-01-Beethoven,%20WID-10841405-N-02-van_Beethoven,%20WID-10841405-N-03-Ludwig_van_Beethoven
2. login in
3. -> error

mkarthik90:
@jdamerow Addressed review comments. We need to create a story for double encoding problems while forwarding the request from login. Temporary fix by removing blank spaces in the url.

PR #149, opened 2016-12-28
story/CCP-187

mkarthik90:


Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptDeleteController.java
```diff
@@ -141,23 +136,4 @@ public ModelAndView confirmlDelete(@PathVariable("id") String id,
...
-            // Need to include command Object
-            model.setViewName("welcome");
-            return model;
-        }
-        conceptManager.deleteConcept(id, principal.getName());
-        model.setViewName("welome");
-        return model;
-    }
-
-}
```
mkarthik90: This method has been introduced due to merge conflicts issue. This method is not called from anywhere. I verified the request mapping.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/ConceptManager.java
```diff
@@ -66,7 +66,7 @@
     @Override
     public ConceptEntry getConceptEntry(String id) {
         ConceptEntry entry = client.getEntry(id);
-        if (entry != null) {
+        if (entry != null && !entry.isDeleted()) {
```
jdamerow: I have my reservations against this approach. The system is built on the assumption that if you give this method a valid id, it'll return a concept (e.g. for synonyms). If you change this it'll have unexpected side effects (e.g. when retrieving synonyms for a concept). If you want to make sure the search doesn't return a deleted concept, you should do that differently. Also, when you said "searching" for a concept, I thought you meant actually searching for the id through the search interface. But you meant retrieving a concept by its ID, right? In this case, CP should probably return a concept but say that it is a deleted concept. Otherwise applications that rely on the deleted concept would break.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java
```diff
@@ -235,9 +239,8 @@ public void deleteIndexes() throws LuceneException {
 
             for (IWordID wordId : wordIdds) {
                 IWord word = dict.getWord(wordId);
-                List<ConceptEntry> entries = databaseClient.getConceptByWordnetId(word.getID().toString());
-                if (entries != null && !entries.isEmpty()) {
-                    logger.debug("Found concept for " + wordId.toString());
+                if (wordnetIdsOfWrappers.contains(word.getID().toString())) {
```
mkarthik90: This was changed because in the previous scenario each concept wrapper will be associated with only one wordnet concept. But from now on since the concepts are getting merged, there are chances three Wordnet concepts can be merged to a single concepts. So in that case we will have three wordnet ids stored in the wordnet variable separated by commas in ConceptEntry. Say now CCP-1 is linked to WID-1, WID-2 and WID-3. When we try to fetch the concept entry based on the wordnet id WID-1, we will not be able to retrieve CCP-1 since we search only by WID-1 but the wordnet id field will contains WID-1, WID-2, WID-3. So the design has been changed to fetch the wordnetIdsOfWrappers and storing it on a Set and by using contains we can make sure whether the wordnet id needs to be indexed or not. If the wordnet id is having a wrapper, there is no need to index it.

mkarthik90:

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java
jdamerow: Hm, the method "getConceptByWordnetId" should always return a concept if there is a concept that wraps the provided wordnet id. ```getConceptByWordnetId``` should be changed to do that then.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/rest/ConceptIDLookup.java
```diff
@@ -88,7 +88,7 @@
 
         XMLConceptMessage msg = messageFactory.createXMLConceptMessage();
 
-        if (entry != null) {
+        if (entry != null && !entry.isDeleted()) {
```
jdamerow: the id lookup should still work for deleted concepts, but when searching it shouldn't.

mkarthik90: @jdamerow  I am not able to understand. This method will be called only through rest interface when the user searches a concept based on id. I am not sure how to differentiate between lookup and searching.

jdamerow:
please resolve conflicts

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/rest/ConceptIDLookup.java
jdamerow: Ups, never mind, sorry. I got confused. Should be correct that way.

mkarthik90:
@jdamerow  Resolved merge conflicts

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/rest/ConceptIDLookup.java
```diff
@@ -91,7 +91,7 @@
 
         IConceptMessage msg = messageFactory.getMessageFactory(acceptHeader).createConceptMessage();
 
-        if (entry != null) {
+        if (entry != null && !entry.isDeleted()) {
```
jdamerow: Wasn't that removed?

jdamerow: Ok, so we seem to go back and forth on this (and I think I was confused at some point too), but if I call .../Concept?id=id_of_deleted_concept, I still want to get the information about the concept I was asking for with the note that the concept was deleted. This happened in the old system, e.g.:
https://diging-dev.asu.edu/conceptpower-latest/rest/Concept?id=CON57c7c59b-a160-4bc1-8f4e-f9a72e14dce5

If Conceptpower suddenly stops returning results for deleted Concepts that might break other applications relying on always getting a response. When a deleted concept was actually wrapped or merged, then it's debatable what should be returned, but it should always return something.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java
```diff
@@ -240,8 +244,7 @@ public void deleteIndexes(String userName) throws LuceneException {
 
             for (IWordID wordId : wordIdds) {
                 IWord word = dict.getWord(wordId);
-                List<ConceptEntry> entries = databaseClient.getConceptByWordnetId(word.getID().toString());
-                if (entries == null || entries.isEmpty()) {
+                if (!wordnetIdsOfWrappers.contains(word.getID().toString())) {
```
jdamerow: still, this means ```getConceptByWordnetId``` needs to be changed, not this code.

mkarthik90:
Addressed review comments.

jdamerow:
ok to test

PR #153, opened 2017-01-25
[CCP-176] Root element is added correctly.

mkarthik90:
Previously pagination was added after root element. Just updated
the root element.


jdamerow:
ok to test

jdamerow:
retest this please

jdamerow:
The tests confirmed what I suspected. You broke any API method that used the ```getAllConceptEntries``` method.

mkarthik90:

PR #154, opened 2017-01-30
[CCP-189] Changed the search field name of equals_to

mkarthik90:
From now on concept search can be done by the predicate equal_to.
Previously the system was using the search term as equals_to but the output
was displaying as equal_to.

### DIFF
--- a/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/constants/SearchFieldNames.java
+++ b/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/constants/SearchFieldNames.java
@@ -2,7 +2,7 @@ package edu.asu.conceptpower.app.constants;
 
 public interface SearchFieldNames {
 
-	public final static String EQUALS_TO = "equals_to";
+    public final static String EQUAL_TO = "equal_to";
 	public final static String SIMILAR_TO = "similar_to";
 	public final static String DESCRIPTION = "description";
 	public final static String WORD = "word";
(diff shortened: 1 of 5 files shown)


jdamerow:
Let's make it ```equal_to``` in all the cases described in the ticket.

mkarthik90:
@jdamerow  Erick has specified about the concept search only. So  I have changed the search predicate in concept search. Is there a different case you are talking about?

mkarthik90:
@jdamerow  Changed all equals predicate to equal_to. While adding a concept, searching a concept. And search result remains unchanged since it was returning equal_to

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/constants/SearchFieldNames.java
```diff
@@ -2,7 +2,7 @@
 
 public interface SearchFieldNames {
 
-	public final static String EQUALS_TO = "equals_to";
+    public final static String EQUAL_TO = "equal_to";
 	public final static String SIMILAR_TO = "similar_to";
 	public final static String DESCRIPTION = "description";
 	public final static String WORD = "word";
```
jdamerow: formatting

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/rest/Concepts.java
```diff
@@ -60,13 +60,13 @@
...
-     *     "description" : "Description",
-     *     "equalsTo" : "equals", 
-     *     "similarTo" : "similar" 
+     *      "wordnetIds":
+     *      "WID-00450866-N-01-pony-trekking,WID-03981924-N-01-pony_cart",
+     *      "synonymids" : "WID-02380464-N-01-polo_pony,WID-04206225-N-03-pony",
+     *      "conceptlist" : "FirstList-1", 
+     *      "type" : "88bea1dc-1443-4296-8315-715c71128b01", 
+     *      "description" : "Description",
+     *      "equal_to" : "equals", "similarTo" : "similar"
```
jdamerow: let's make ```equals``` and ```similar``` actual VIAF uris. also line break is missing.

jdamerow:
ok to test

jdamerow:
Should be ok after those two things

mkarthik90:

PR #156, opened 2017-03-07
Story/CCP-202

mkarthik90:


mkarthik90:
I have included servlet context in integration test because we need to index the concepts before running the tests. And lucene utility is available through servlet context. For now, I have placed the database files in the workspace, but we can keep it in a different location and provide that location in ${db.path}. 

Also I have included one integration test for welcome UI page, In future, we can add up integration tests for UI pages as well.

Conceptpower+Spring/pom.xml
```diff
@@ -204,12 +223,12 @@
...
-			<groupId>javax.servlet</groupId>
-			<artifactId>servlet-api</artifactId>
-			<version>2.5</version>
-			<scope>provided</scope>
-		</dependency>
+        <dependency>
+            <groupId>javax.servlet</groupId>
+            <artifactId>javax.servlet-api</artifactId>
+            <version>3.1.0</version>
+        </dependency>
```
jdamerow: do you need a newer servlet version for this? we're still running tomcat 6 in production which doesn't support this.

Conceptpower+Spring/src/integration-test/java/edu/asu/conceptpower/IntegrationTest.java
```diff
@@ -0,0 +1,55 @@
...
+    protected Principal principal = new Principal() {
+        public String getName() {
+            return "admin";
+        }
+    };
+
+    @Before
+    public void setup() throws Exception {
+
+        this.mockMvc = MockMvcBuilders.webAppContextSetup(this.wac).build();
```
jdamerow: before running the indexer, you should add the database files. 
- create a smaller version of our production database (maybe just take some example from production and add them to a fresh install, then use that database).
- add database to git
- copy db into place before indexing

Conceptpower+Spring/src/integration-test/java/edu/asu/conceptpower/rest/ConceptLookupIT.java
```diff
@@ -0,0 +1,100 @@
...
+                .andExpect(jsonPath("$.conceptEntries[3].description",
+                        is("someone who has exceptional intellectual ability and originality; \"Mozart was a child genius\"; \"he's smart but he's no Einstein\"")))
+                .andExpect(jsonPath("$.conceptEntries[3].conceptList", is("WordNet")))
+                .andExpect(jsonPath("$.conceptEntries[3].deleted", is(false)))
+                .andExpect(jsonPath("$.conceptEntries[3].concept_uri",
+                        is("http://www.digitalhps.org/concepts/WID-10126926-N-05-Einstein")))
+                .andExpect(jsonPath("$.conceptEntries[3].synonym_ids",
+                        is("WID-10126926-N-01-genius,WID-10126926-N-02-mastermind,WID-10126926-N-03-brain,WID-10126926-N-04-brainiac,")))
+                .andExpect(jsonPath("$.conceptEntries[3].wordnet_id", is("WID-10126926-N-05-Einstein")))
+                .andExpect(status().isOk());
```
jdamerow: Can't you compare the output to a string stored in a file? Would make it easier to maintain and easier to read.

jdamerow: This is true for all tests.

Conceptpower+Spring/src/test/resources/wordnet-config-test.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<beans xmlns="http://www.springframework.org/schema/beans"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd">
+
+
+	<bean class="edu.asu.conceptpower.app.wordnet.WordNetConfiguration">
+        <!-- TODO -->
+		<property name="wordnetPath" value="/usr/local/Cellar/wordnet/3.1" />
```
jdamerow: this cannot be hard coded

jdamerow:
what about xml output?

Conceptpower+Spring/pom.xml
```diff
@@ -29,6 +29,8 @@
 		<lucene.index>/Users/karthikeyanmohan/Software/Conceptpower/indexFiles</lucene.index>
 		<pullrequest></pullrequest>
 		<debug_level>debug</debug_level>
+        <skip.integration.tests>false</skip.integration.tests>
+        <skip.unit.tests>false</skip.unit.tests>
```
mkarthik90: Will format entire file before merging to develop

mkarthik90:

Conceptpower+Spring/src/integration-test/java/edu/asu/conceptpower/IntegrationTest.java
jdamerow: are you copying the db anywhere during setup?

Conceptpower+Spring/src/test/resources/root-context-test.xml
```diff
@@ -0,0 +1,70 @@
...
+	<import resource="db4o-config-test.xml" />
+	<import resource="xml-config-test.xml" />
+	<context:component-scan base-package="edu.asu.conceptpower.app" />
+    
+    <task:executor id="indexExecutor" pool-size="1" />      
+    <task:annotation-driven executor="indexExecutor"/>
+    <aop:aspectj-autoproxy proxy-target-class="true" />
+
+	<bean id="luceneDAO" class="edu.asu.conceptpower.app.lucene.impl.LuceneDAO" init-method="init">
+		<property name="dbPath" value="/Users/karthikeyanmohan/Software/conceptpower-test/lucene.db" />
```
jdamerow: Can't be hardcoded

Conceptpower+Spring/src/integration-test/java/edu/asu/conceptpower/IntegrationTest.java
jdamerow: Ok, I think I see. But I don't think this will work. I can't necessarily know ahead of time where the folder will be that will contain the dbs (it's up to Jenkins to checkout the project). Let's copy the files to a location specified through dbpath. Also, the db files should not be in WEB-INF resources as that would add them to the generated war files, which we don't want because they artificially increase the war file size. Let's put them somewhere else (test resource folder should work).

mkarthik90:

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/DownloadFilesUtility.java
```diff
@@ -0,0 +1,80 @@
...
+    @Value("${dbTestPath}")
+    private String toLocation;
+
+    public void setUp() throws IOException {
+
+        ReadableByteChannel rbc = null;
+        FileOutputStream fos = null;
+        URL website = null;
+        try {
+            website = new URL("https://www.dropbox.com/s/t2g2kclkwglv2lp/conceptLists.db?raw=1");
```
jdamerow: this should be configurable in case the url changes

jdamerow:
Also, didn't we say the integration tests should be developed off the master branch (latest release) so that we can make sure the new changes don't break anything?

PR #163, opened 2017-03-27
[CCP-218] [CCP-217]Changes for displaying merged ids and merged concepts.

mkarthik90:
When user searches with the word net id that has been merged, the merged
concept will be displayed, Previously the wrapper of the word net was displayed but from
now the merged concept itself will be displayed.

Also the RESTFul apis will display the merged ids of the concepts.


jdamerow:
ok to test

jdamerow:
run integration tests please

PR #166, opened 2017-03-29
[CCP-210] Concept search requires atleast one search parameter

mkarthik90:
From now on concept search rest api requires atleast one search parameter. If no
parameter is provided, the system returns the error message "No search condition
present" and return 400 http status code.

Previously the system went into 500 error when no parameter is provided.

### DIFF
--- a/Conceptpower+Spring/src/integration-test/resources/output/conceptSearchNoSearchParameters.json
+++ b/Conceptpower+Spring/src/integration-test/resources/output/conceptSearchNoSearchParameters.json
@@ -1,0 +1,1 @@
+[{"message":"No valid search parameters present."}]
--- a/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/rest/ConceptSearchParameterValidator.java
+++ b/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/rest/ConceptSearchParameterValidator.java
@@ -21,7 +21,7 @@ import edu.asu.conceptpower.app.core.POS;
 @Component
 public class ConceptSearchParameterValidator implements Validator {
 
-    private static final Logger logger = LoggerFactory.getLogger(ConceptSearchParameterValidator.class);
+    private final Logger logger = LoggerFactory.getLogger(ConceptSearchParameterValidator.class);
 
     @Override
     public boolean supports(Class<?> clazz) {
@@ -33,7 +33,7 @@ public class ConceptSearchParameterValidator implements Validator {
         ConceptSearchParameters conceptSearchParameter = (ConceptSearchParameters) target;
 
         if (!hasAtLeastOneNonEmpty(conceptSearchParameter)) {
-            errors.reject("validSearch", "No valid search parameters present.");
+            errors.reject("invalidSearch", "No valid search parameters present.");
         }
         if (conceptSearchParameter.getPos() != null && !POS.posValues.contains(conceptSearchParameter.getPos().toLowerCase())) {
             errors.reject("pos", "Please enter correct pos value.");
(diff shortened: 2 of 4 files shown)


Conceptpower+Spring/src/main/java/edu/asu/conceptpower/rest/ConceptSearchParameterValidator.java
```diff
@@ -36,4 +46,19 @@ public void validate(Object target, Errors errors) {
         }
     }
 
+    private boolean hasAtLeastOneNonEmpty(ConceptSearchParameters conceptSearchParameters) {
+        Field[] fields = ConceptSearchParameters.class.getDeclaredFields();
+        for (Field field : fields) {
+            field.setAccessible(true);
+            try {
+                if (field.get(conceptSearchParameters) != null) {
```
mkarthik90: Let me know if I need to check for empty spaces as well.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/rest/ConceptSearchParameterValidator.java
```diff
@@ -17,6 +21,8 @@
 @Component
 public class ConceptSearchParameterValidator implements Validator {
 
+    private static final Logger logger = LoggerFactory.getLogger(ConceptSearchParameterValidator.class);
```
jdamerow: doesn't need to be static

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/rest/ConceptSearchParameterValidator.java
```diff
@@ -36,4 +46,19 @@ public void validate(Object target, Errors errors) {
...
+    private boolean hasAtLeastOneNonEmpty(ConceptSearchParameters conceptSearchParameters) {
+        Field[] fields = ConceptSearchParameters.class.getDeclaredFields();
+        for (Field field : fields) {
+            field.setAccessible(true);
+            try {
+                if (field.get(conceptSearchParameters) != null) {
+                    return true;
+                }
+            } catch (IllegalArgumentException | IllegalAccessException e) {
+                logger.error(e.getMessage());
```
jdamerow: always log the whole exception

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/rest/ConceptSearchParameterValidator.java
```diff
@@ -25,6 +31,10 @@ public boolean supports(Class<?> clazz) {
     @Override
     public void validate(Object target, Errors errors) {
         ConceptSearchParameters conceptSearchParameter = (ConceptSearchParameters) target;
+
+        if (!hasAtLeastOneNonEmpty(conceptSearchParameter)) {
+            errors.reject("validSearch", "No valid search parameters present.");
```
jdamerow: shouldn't that be "invalid" search?

jdamerow: as message key I mean

jdamerow:
ok to test

jdamerow:
run integration tests please

Conceptpower+Spring/src/integration-test/resources/output/conceptSearchNoSearchParameters.json
```diff
@@ -0,0 +1 @@
+[{"message":"No valid search parameters present."}]
```
jdamerow: why is this a list? could there be more than one error message?

mkarthik90: Yes, there could be multiple error messages. Say for example if we are having an invalid pos and an invalid operator we return both the error message to user. So that's the reason I had it as an array of error messages.

jdamerow: makes sense

PR #171, opened 2017-04-17
[CCP-227] Changes for deleting word net concepts.

mkarthik90:
Currently word net concepts are not deleted when a wrapper is updated
with a word net id. But from now on when a word net id is added to wrapper
that word net concept will be deleted.


jdamerow:
ok to test

jdamerow:
the problem exists the other way around too though. If a wordnet id is deleted from a concept, the concept is just gone (it won't show up as wordnet concept in the search anymore).

mkarthik90:
reopening to force push a commit

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/IConceptEditService.java
```diff
@@ -0,0 +1,23 @@
...
+     * This method edits the concept.
+     * 
+     * @param conceptEntry
+     * @param conceptEditBean
+     * @param userName
+     * @throws IllegalAccessException
+     * @throws LuceneException
+     * @throws IndexerRunningException
+     */
+    public void editConcepts(ConceptEntry conceptEntry, ConceptEditBean conceptEditBean, String userName)
```
jdamerow: bad javadoc, you only say what the method name already tells me. what does this method do?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptEditService.java
```diff
@@ -0,0 +1,74 @@
...
+import edu.asu.conceptpower.core.ConceptEntry;
+
+@Service
+public class ConceptEditService implements IConceptEditService {
+
+    @Autowired
+    private IConceptManager conceptManager;
+
+    /**
+     * This method adds all the wordnet concepts that are removed from wordnet
```
jdamerow: add all wordnet concepts to what?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptEditService.java
```diff
@@ -0,0 +1,74 @@
...
+     * id field of wrapper. The next step is to remove all the wordnet concepts
+     * that are added to the wordnet id field of the wrapper. Finally the
+     * concept is updated.
```
jdamerow: what does updated mean here? saved in database, values are updated, index is updated?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptEditService.java
```diff
@@ -0,0 +1,74 @@
...
+     * 
+     * 
+     * If conceptEntry or conceptEditBean is null, this method just returns
+     * without editing the concept.
+     *
```
jdamerow: what do the parameters do?

mkarthik90:
@jdamerow Updated java doc.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/IConceptEditService.java
```diff
@@ -0,0 +1,27 @@
...
+     * 
+     * Apart from updating the concept, this method adds and removes the wordnet
+     * entries from the lucene index and database.
+     * 
+     * @param conceptEntry
+     * @param conceptEditBean
+     * @param userName
+     * @throws IllegalAccessException
+     * @throws LuceneException
+     * @throws IndexerRunningException
```
jdamerow: description of parameters still missing

mkarthik90:

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/IConceptEditService.java
```diff
@@ -0,0 +1,30 @@
...
+     * 
+     * Apart from updating the concept, this method adds and removes the wordnet
+     * entries from the lucene index and database.
+     * 
+     * @param conceptEntry
+     *            - concept entry to be updated in the database and lucene
+     * @param conceptEditBean
+     *            - contains the details of wordnet ids that are removed and
+     *            added to concept entry
+     * @param userName
```
jdamerow: what about this one?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptEditService.java
```diff
@@ -0,0 +1,79 @@
...
+            throws IllegalAccessException, LuceneException, IndexerRunningException {
+
+        if (conceptEntry == null || conceptEditBean == null) {
+            return;
+        }
+
+        Set<String> oldWordnetIds = new HashSet<String>(
+                Arrays.asList(conceptEditBean.getExistingWordnetIds().split(",")));
+        Set<String> updatedWordnetIds = new HashSet<String>(Arrays.asList(conceptEditBean.getWordnetIds().split(",")));
+        oldWordnetIds.removeAll(updatedWordnetIds);
```
jdamerow: since the conceptEntry already contains the updated information, and the conceptEditBean is only passed to get two lists of wordnet ids, just pass the wordnet id lists directly. The way it is right now is just confusing as it would imply that the method copies the updated data from the editbean to the concept entry, but it doesn't.

mkarthik90:

jdamerow:
please resolve conflicts

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/ConceptManager.java
```diff
@@ -109,6 +116,10 @@ public ConceptEntry getWordnetConceptEntry(String wordnetId) throws LuceneExcept
 
     @Override
     public ConceptEntry getLocalConceptEntry(String conceptId) {
```
jdamerow: it seems like this method is only used once (for deleting a concept after a merge). and only to avoid searching in merged ids when retrieving a concept. Why don't you just retrieve the concepts to be deleted before the merge using the getConceptEntry method and then delete those concepts after the merge has been successful?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/ConceptManager.java
```diff
@@ -572,4 +582,11 @@ private int getConceptsCountForConceptList(String listName) {
         alternativeIdService.addAlternativeIds(conceptEntries);
         return conceptEntries;
     }
+
+    @Override
+    public void addWordnetConceptEntry(ConceptEntry entry, String userName)
```
jdamerow: why does this need its own method? why not setting the list before adding the concept entry?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DatabaseClient.java
```diff
@@ -77,6 +59,18 @@ public ConceptEntry getEntry(String id) {
     }
 
     @Override
+    public ConceptEntry getWrapperEntryByWordnetId(String wordnetId) {
```
jdamerow: this and the next method seem to try to do the same thing. As wordnet ids can be multiple and therefore the ```getConceptByWordnetId``` doesn't really work, those two methods should be consolidated.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DatabaseClient.java
```diff
@@ -377,6 +371,9 @@ public void store(Object element, String databasename) {
     public void update(ConceptEntry entry, String databasename) {
         if (databasename.equals(DBNames.DICTIONARY_DB)) {
             ConceptEntry toBeUpdated = getEntry(entry.getId());
+            if (toBeUpdated == null) {
+                toBeUpdated = getWrapperEntryByWordnetId(entry.getId());
+            }
```
jdamerow: can that ever be the case? this again is business logic in the database layer.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/objectdb/ConceptDBManager.java
```diff
@@ -2,18 +2,12 @@
```
jdamerow: let's remove that class

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/ConceptManager.java
mkarthik90: Yes, I thought about that, but the method addConceptListEntry(ConceptEntry entry, String userName) has an if condition that checks whether concept list is "wordnet", if so it throws DictionaryModificationException. Maybe I could reuse this method if I remove this if condition. Let me know if I can remove this and reuse this method? Will update the test cases as well.

jdamerow: now I'm really confused. This method actually doesn't add a concept entry. It's only adding it to the index. If anything there should be a method to add a concept to the index. The list should be set before this is done.

jdamerow: and similarly the method ```addWordNetConcepts``` in ```ConceptEditService``` needs to be named in a way that makes clear it's not adding a concept. It just updates the index.

mkarthik90:

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/ConceptManager.java
```diff
@@ -572,4 +574,10 @@ private int getConceptsCountForConceptList(String listName) {
         alternativeIdService.addAlternativeIds(conceptEntries);
         return conceptEntries;
     }
+
+    @Override
+    public void updateWordnetConceptInIndex(ConceptEntry entry, String userName)
```
jdamerow: just call it ```updateIndex``` since this would work for any entry. If the concept already exist, will the old index entry be updated? Or would there be a second entry?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DatabaseClient.java
```diff
@@ -77,12 +60,10 @@ public ConceptEntry getEntry(String id) {
...
-    public List<ConceptEntry> getConceptByWordnetId(String wordnetId) {
-        ConceptEntry entry = new ConceptEntry();
-        entry.setWordnetId(wordnetId);
-
-        ObjectSet<ConceptEntry> entries = dictionaryClient.queryByExample(entry);
-        return entries;
+    public List<ConceptEntry> getWrapperEntryByWordnetId(String wordnetId) {
+        ConceptEntry exampleEntry = new ConceptEntry();
+        exampleEntry.setWordnetId(wordnetId);
+        return Arrays.asList(getEntriesByFieldContains("wordnetid", wordnetId));
```
jdamerow: what is the exampleentry for?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/ConceptManager.java
```diff
@@ -576,8 +575,13 @@ private int getConceptsCountForConceptList(String listName) {
...
     @Override
-    public void updateWordnetConceptInIndex(ConceptEntry entry, String userName)
+    public void updateIndex(ConceptEntry entry, String userName)
             throws IllegalAccessException, LuceneException, IndexerRunningException {
-        indexService.insertConcept(entry, userName);
+        indexService.updateConceptEntry(entry, userName);
+    }
+
+    @Override
+    public void deleteIndex(String id, String userName) throws LuceneException, IndexerRunningException {
```
jdamerow: given that this does not delete the whole index, i'd call it deleteFromIndex

jdamerow:
ok to test

mkarthik90:
Changed the method name to deleteFromIndex.

jdamerow:
run integration tests please

PR #172, opened 2017-04-17
[CCP-222] Deletion is done before merging

mkarthik90:
Deletion is done before merging, because once the concepts are merged, the change made
in the getConceptEntry in ConceptManager always fetches the merged concept. Because of
this instead of fetching the original concept and setting it to deleted, we set the delete
flag on the merged concept.

### DIFF
--- a/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/ConceptManager.java
+++ b/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/ConceptManager.java
@@ -107,6 +107,11 @@ public class ConceptManager implements IConceptManager {
         return entry;
     }
 
+    @Override
+    public ConceptEntry getLocalConceptEntry(String conceptId) {
+        return client.getEntry(conceptId);
+    }
+
     /*
      * (non-Javadoc)
      * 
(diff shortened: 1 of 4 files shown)


Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/ConceptManager.java
```diff
@@ -524,6 +524,18 @@ public void deleteConcept(String id, String userName) throws LuceneException, In
...
+    public void deleteNonMergedConcept(String id, String userName) throws LuceneException, IndexerRunningException {
+        ConceptEntry concept = client.getEntry(id);
+        concept.setDeleted(true);
+        ChangeEvent changeEvent = new ChangeEvent();
+        changeEvent.setType(ChangeEventTypes.DELETION);
+        changeEvent.setDate(new Date());
+        changeEvent.setUserName(userName);
+        concept.addNewChangeEvent(changeEvent);
+        client.update(concept, DBNames.DICTIONARY_DB);
+        indexService.deleteById(concept.getId(), userName);
```
jdamerow: the previous method does pretty much exactly the same thing. just have one method and give it the concept to delete.

mkarthik90:

jdamerow:
the code doesn't seem to have any changes to the order in which concepts are deleted and merged. Is that correct?

mkarthik90:
@jdamerow Initially I moved the deletion part before the merging part, but then I realized that it is not a good practice to delete the concept and create the merged concepts. I forgot to update the comments. 
To answer your comment, there are **no changes** to the order of merging and deletion in concept merge service.

jdamerow:
ok, cool. that sounds right.

jdamerow:
ok to test

PR #176, opened 2017-05-03
Story/CCP-212

mkarthik90:
Pagination and sorting changes for concept search screen. 
Will include test case once review comments are addressed.


mkarthik90:
run integration tests please

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java
```diff
@@ -206,6 +215,15 @@ public void insertConcept(ConceptEntry entry, String userName) throws LuceneExce
                         doc.add(new StringField(searchFieldAnnotation.lucenefieldName() + LuceneFieldNames.UNTOKENIZED_SUFFIX,
                                 String.valueOf(contentOfField), Field.Store.YES));
                     }
+
+                    if (searchFieldAnnotation.isSortAllowed()) {
+                        String value = StringEscapeUtils
+                                .unescapeHtml4(String.valueOf(contentOfField).replaceAll("\\s", "")).trim()
+                                .toLowerCase();
```
jdamerow: Couple of things here:
- it seems like this line is used more often; make that a method to call
- why do you remove whitespaces? 'hello world' is very different 'helloworld' and should be sorted to different positions
- html contents should be removed for sorting (the users only sees 'pony' and '**pony**' and would see why they would be on different positions.
- non-ascii characters like umlauts, accents, etc. should probably be turned into ascii for sorting (ö -> o, é -> e, etc.)

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptSearchController.java
```diff
@@ -53,28 +55,43 @@
...
+            @RequestParam(defaultValue = IConceptDBManager.ASCENDING + "") String sortDir,
+            @RequestParam(defaultValue = "word") String sortColumn,
+            @RequestParam(required = false, defaultValue = "") String conceptIdsToMerge,
             @Validated @ModelAttribute("conceptSearchBean") ConceptSearchBean conceptSearchBean, BindingResult results)
                     throws LuceneException, IllegalAccessException {
         if (results.hasErrors()) {
             return "conceptsearch";
         }
+        
+        if (conceptIdsToMerge != null && !conceptIdsToMerge.equals("")) {
```
jdamerow: I don't think this check is really necessary. If the attribute is null, it's the same as when it's not set I believe.

mkarthik90:
@jdamerow  Addressed review comments. For replacing the non-ascii characters with ascii characters, I replaced with empty string and sorting seems to work fine. Let me know if that's fine.

mkarthik90:

jdamerow:
ok to test

jdamerow:
Couple of things:
- default sorting has to be by relevancy with exact matches first
- table headings need to have a little icon that shows the current sorting and direction (e.g. up/down arrow for ascending/descending, both for unsorted columns that can be sorted)

mkarthik90:

jdamerow:
Ok, let's put some space between icon and column title so that it doesn't look so squished and we should be good to go.

mkarthik90:
@jdamerow  Addressed review comment. Added spacing.

PR #177, opened 2017-05-04
Bug/CCP-226

mkarthik90:
Opened for testing integration tests


mkarthik90:
run integration tests please

mkarthik90:
@jdamerow  Changed the phrase query to term query because phrase query not fetching the values. And also while indexing the stop words shouldn't be removed to set the stop words to empty in Standard analyzer.

jdamerow:
from the ticket: "let's make this optional; in rest interface there should be a flag to set, and gui interface there should be a checkbox or something."

jdamerow:
ok to test

jdamerow:
There definitely needs to be a label for what this checkbox does, how would anyone know otherwise? Also, I'd put the checkbox under the input field.

mkarthik90:

jdamerow:
the search should be "or", eg.: 'einstein' in word OR 'einstein' in description

PR #180, opened 2017-05-05
[CCP-229] Changes for displaying the concept after creating it.

mkarthik90:


jdamerow:
Instead of making all those changes, you could have just make another call to the concept manager after you got the concept id. Would have been easier, but this should work too.

jdamerow:
ok to test

mkarthik90:
@jdamerow  Yes I thought I could do that but since that would be one more database call,  I decided to do this way. Fixed the test cases.

PR #181, opened 2017-05-05
[CCP-220] Used unescape from apache.

mkarthik90:
This will unescape the characters in all uris. But if the description field contains
a \", then this might not work. Not sure if the rest api call can contain a \" character.


mkarthik90:
@jdamerow This will unescape the characters in all uris. But if the description field contains
a \", then this will not work. Not sure if the rest api call can contain a \" character.

jdamerow:
why don't you just escape the fields that can contain uris (equals, similar to, etc)

jdamerow:
ok to test

jdamerow:
Jenkins, run integration tests please

PR #183, opened 2017-07-17
[CCP-233] lowercase indexed fields for case insensitive search

jdamerow:


jdamerow:
run integration tests please

PR #186, opened 2018-02-26
Bug/ccp 234

susmitha8:
Hyphenated search term handled and the term is found

### DIFF
--- a/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java
+++ b/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java
@@ -622,7 +622,9 @@ public class LuceneUtility implements ILuceneUtility {
 
     private void buildTokenizedOrWildCardQuery(LuceneField luceneFieldAnnotation, String searchString,
             BooleanQuery.Builder tokenizedQueryBuilder) {
-        for (String searchValue : searchString.split(" ")) {
+        String hypSearchStr = new String(searchString);
+        hypSearchStr = searchString.replace('-', ' ');
+        for (String searchValue : hypSearchStr.split(" ")) {
             if (luceneFieldAnnotation.isWildCardSearchEnabled()) {
                 createWildCardSearchQuery(luceneFieldAnnotation, searchValue, tokenizedQueryBuilder, Occur.MUST);
             } else {
--- a/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java
+++ b/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java
@@ -624,15 +624,17 @@ public class LuceneUtility implements ILuceneUtility {
             BooleanQuery.Builder tokenizedQueryBuilder) {
         String hypSearchStr = new String(searchString);
         hypSearchStr = searchString.replace('-', ' ');
-        for (String searchValue : hypSearchStr.split(" ")) {
             if (luceneFieldAnnotation.isWildCardSearchEnabled()) {
+                for (String searchValue : hypSearchStr.split(" ")) {
                 createWildCardSearchQuery(luceneFieldAnnotation, searchValue, tokenizedQueryBuilder, Occur.MUST);
+                }
             } else {
+                for (String searchValue : hypSearchStr.split(" ")) {
                 tokenizedQueryBuilder.add(new PhraseQuery(luceneFieldAnnotation.lucenefieldName(), searchValue),
                         Occur.MUST);
+                }
             }
         }
-    }
 
     /**
      * This method adds the wild card query to the query builder when the
--- a/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptSearchController.java
+++ b/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptSearchController.java
@@ -87,6 +87,8 @@ public class ConceptSearchController {
         if (results.hasErrors()) {
             return "conceptsearch";
         }
+        String str = conceptSearchBean.getWord().trim();
+        conceptSearchBean.setWord(str);
         
         if (conceptIdsToMerge != null) {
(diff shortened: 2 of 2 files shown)


Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptSearchController.java
```diff
@@ -87,6 +87,8 @@ public String search(HttpServletRequest req, ModelMap model, @RequestParam(defau
         if (results.hasErrors()) {
             return "conceptsearch";
         }
+        String str = conceptSearchBean.getWord().trim();
+        conceptSearchBean.setWord(str);
```
chiraag-subramanian: The lines 90 and 91 can be merged into a single statement: conceptSearchBean.setWord(conceptSearchBean.getWord().trim())

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java
```diff
@@ -622,7 +622,9 @@ private void buildQuery(BooleanClause.Occur occur, PerFieldAnalyzerWrapper perFi
 
     private void buildTokenizedOrWildCardQuery(LuceneField luceneFieldAnnotation, String searchString,
             BooleanQuery.Builder tokenizedQueryBuilder) {
-        for (String searchValue : searchString.split(" ")) {
+        String hypSearchStr = new String(searchString);
+        hypSearchStr = searchString.replace('-', ' ');
+        for (String searchValue : hypSearchStr.split(" ")) {
             if (luceneFieldAnnotation.isWildCardSearchEnabled()) {
                 createWildCardSearchQuery(luceneFieldAnnotation, searchValue, tokenizedQueryBuilder, Occur.MUST);
```
chiraag-subramanian: Does the value of luceneFieldAnnotation.isWildCardSearchEnabled() change for every iteration? 

If not, its better to write the if-else condition and then include the iteration call inside them: 
String splitHypSearchStr =  hypSearchStr.split(" ");
if(luceneFieldAnnotation.isWildCardSearchEnabled()){
     // iterate splitHypSearchStr and call the method createWildCardSearchQuery
}
else{
  // iiterate splitHypSearchStr and call the method tokenizedQueryBuilder.add
}

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptSearchController.java
susmitha8: Yes, I'll merge these into single statement

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java
susmitha8: The value of luceneFieldAnnotation.isWildCardSearchEnabled() doesn't change for every iteration of hypSearchStr. But instead of having 2 iterators isn't it better to have just one iterator outside the if else?

chiraag-subramanian: Given that the value of  luceneFieldAnnotation.isWildCardSearchEnabled()  doesn't change for every iteration, iI feel its good to have the iterations inside both if and else loop, instead of checking the conditions (if/else) for every iteration.  If you feel that the code will look cluttered due to the for-loops inside if and else, consider : 
1) 
https://stackoverflow.com/questions/27534684/good-method-to-call-method-on-each-object-using-stream-api
2) 
https://stackoverflow.com/questions/36508530/converting-array-iteration-to-lambda-expression

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ConceptSearchController.java
Mihika0409: lines 90 and 91 can be merged

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java
```diff
@@ -622,15 +622,19 @@ private void buildQuery(BooleanClause.Occur occur, PerFieldAnalyzerWrapper perFi
 
     private void buildTokenizedOrWildCardQuery(LuceneField luceneFieldAnnotation, String searchString,
             BooleanQuery.Builder tokenizedQueryBuilder) {
-        for (String searchValue : searchString.split(" ")) {
+        String hypSearchStr = new String(searchString);
+        hypSearchStr = searchString.replace('-', ' ');
```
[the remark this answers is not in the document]
abhikrish503501: Lines 625 and 626 can be merged into single statement.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java
```diff
@@ -622,15 +622,19 @@ private void buildQuery(BooleanClause.Occur occur, PerFieldAnalyzerWrapper perFi
...
             if (luceneFieldAnnotation.isWildCardSearchEnabled()) {
+                for (String searchValue : hypSearchStr.split(" ")) {
                 createWildCardSearchQuery(luceneFieldAnnotation, searchValue, tokenizedQueryBuilder, Occur.MUST);
+                }
             } else {
+                for (String searchValue : hypSearchStr.split(" ")) {
                 tokenizedQueryBuilder.add(new PhraseQuery(luceneFieldAnnotation.lucenefieldName(), searchValue),
                         Occur.MUST);
+                }
```
jdamerow: this is not the solution for the problem; we will have the same problem most likely if there are other word separating characters (e.g. :, ~, etc). You need to make sure the search terms are normalized/analyzed before the query is created.

susmitha8: This change was merged to develop as part of Normalized value search change

PR #191, opened 2018-04-24
wildcard search with ascii characters

susmitha8:


Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java
```diff
@@ -157,11 +158,15 @@ public void init() throws LuceneException, IOException {
                 addTokenFilter("lowercase").build();
         lucenePath = env.getProperty("lucenePath");
         numberOfResults = Integer.parseInt(env.getProperty("numberOfLuceneResults"));
+        customAnalyzer2 = CustomAnalyzer.builder().withTokenizer("keyword").addTokenFilter("lowercase").build();
+        Map<String, Analyzer> perFieldAnalyzer = new HashMap<>();
+        perFieldAnalyzer.put("wordtest", customAnalyzer2);
```
jdamerow: the field need to be implemented like all the others; if they are wildcard searchable, there need to be a proper field name for the fields and all of them need to be read our here. 'wordtest' was just for testing (as the name suggests).

PR #199, opened 2019-09-25
[CCP-195] Added numberOfPages to the return object on search

kkrish39:


Conceptpower+Spring/src/main/java/edu/asu/conceptpower/rest/ConceptSearch.java
```diff
@@ -133,9 +133,14 @@ private void initBinder(WebDataBinder binder) {
 
         ConceptEntry[] searchResults = null;
         int totalNumberOfRecords = 0;
+        int numberOfPages = 0;
+        numberOfRecordsPerPage = (conceptSearchParameters.getNumber_of_records_per_page()!= null
```
jdamerow, line 137: why do you overwrite this here? this is a default value that needs to stay the default.

jdamerow (changes requested):

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/rest/ConceptSearch.java
kkrish39: There was a field in ConceptSearchParameter class as number_of_records_per_page. I though we will extend in future for custom number of records per page. To handle that case I've added that.

jdamerow: you are overwriting the default for everybody else.

PR #201, opened 2019-10-24
Feature/CCP 253

kkrish39:


kkrish39:
home.jsp contains the html indentation fix along with the logic addition.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
```diff
@@ -30,7 +32,20 @@ public void init() {
...
+            @Override
+            public boolean match(ReviewRequest review) {
+                return review.getConceptId().equals(reviewRequest.getConceptId());
+            }
+            
+        });
+        //Deleting the reviewRequest if already present and Adding it back with updated Request data
+        if(isReviewRequestPresent.size() !=0) {
+            client.delete(isReviewRequestPresent.get(0));
+        }
```
jdamerow: why don't you update the existing one?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
```diff
@@ -48,5 +63,49 @@ public ReviewRequest getReviewRequestForConcept(String conceptId){
...
+        
+        requestToBeUpdated.setStatus(ReviewStatus.OPENED);
+        requestToBeUpdated.setResolver(null);
+        requestToBeUpdated.setResolvingComment(null);
+        
+        client.store(requestToBeUpdated);
+        client.commit();
+        
+        return requestToBeUpdated;
+    }
```
jdamerow, line 60: All the business logic should be done in the manager or you have to rewrite it all when we switch the db backend.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/core/ReviewRequest.java
```diff
@@ -49,5 +50,13 @@ public String getResolver() {
...
         this.resolver = resolver;
     }  
+    
+    public String getResolvingComment() {
+        return resolvingComment;
+    }
+    
+    public void setResolvingComment(String resolvingComment) {
+        this.resolvingComment = resolvingComment;
+    }
```
jdamerow, line 85: I would allow for a list of comments since it's likely we want to allow discussions at some point.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ReviewRequestController.java
```diff
@@ -21,12 +21,28 @@
...
         
         requestsMgr.addReviewRequest(reviewRequest);
         
         return reviewRequest;
     }
+    
+    @RequestMapping(value = "/auth/request/resolve", method = RequestMethod.POST )
+    public @ResponseBody ReviewRequest resolveRequest(@ModelAttribute(value="reviewRequest") ReviewRequest reviewRequest,Principal principal) {
+        reviewRequest.setStatus(ReviewStatus.RESOLVED);
+        reviewRequest.setResolver(principal.getName());
```
jdamerow: what about the comment?

Conceptpower+Spring/src/main/webapp/WEB-INF/views/home.jsp
```diff
@@ -191,21 +191,26 @@ function prepareMergeConcept(conceptId) {
 function createWrapper(wrapperId) {
   window.location = '${pageContext.servletContext.contextPath}/auth/conceptlist/addconceptwrapper?wrapperId=' + wrapperId;
 }
+
 $(document).ready(function(){
-    $(".fa-exclamation-triangle").click(function() {
+    $(document).on("click", ".fa-exclamation-triangle", function() {
```
jdamerow: why did you change that?

Conceptpower+Spring/src/main/webapp/WEB-INF/views/home.jsp
```diff
@@ -191,21 +191,26 @@ function prepareMergeConcept(conceptId) {
...
 		$("#fetchRequests").val($(this).data("request"));
+    	$("#conceptId").val($(this).data("concept-id"));
+    	$('#resolveCommentError').hide();
 		$("#requestBox").show();    
 	});
-    $(".fa-comment").click(function() {
+    
+    $(document).on("click", ".fa-comment", function() {
```
jdamerow: and this?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
```diff
@@ -48,5 +63,49 @@ public ReviewRequest getReviewRequestForConcept(String conceptId){
...
+    
+    @Override
+    public ReviewRequest updateReviewRequest(ReviewRequest reviewRequest) {
+        
+        ReviewRequest requestToBeUpdated = client.query(new Predicate<ReviewRequest>() {
+            private static final long serialVersionUID = 6495914730735826451L;
+
+            @Override
+            public boolean match(ReviewRequest review) {
+                return review.getConceptId().equals(reviewRequest.getConceptId());
```
jdamerow: you can just query by example, no predicate needed (it's usually really slow). you should also be able to index by concept id, which should speed it up.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
```diff
@@ -48,5 +63,49 @@ public ReviewRequest getReviewRequestForConcept(String conceptId){
...
+    }
+    
+    @Override
+    public ReviewRequest reopenReviewRequest(String conceptId) {
+        ReviewRequest requestToBeUpdated = client.query(new Predicate<ReviewRequest>() {
+            private static final long serialVersionUID = 6495914730735826451L;
+
+            @Override
+            public boolean match(ReviewRequest review) {
+                return review.getConceptId().equals(conceptId);
```
jdamerow: same as above

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
```diff
@@ -48,5 +63,49 @@ public ReviewRequest getReviewRequestForConcept(String conceptId){
...
+            }
+            
+        }).get(0);
+        
+        requestToBeUpdated.setStatus(ReviewStatus.OPENED);
+        requestToBeUpdated.setResolver(null);
+        requestToBeUpdated.setResolvingComment(null);
```
jdamerow: this is business logic and belongs into the manager class.

jdamerow (changes requested):
looks like i never published my review

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/db/DBRequestClientTest.java
```diff
@@ -50,4 +77,32 @@ public void getReviewRequestTest() {
         assertEquals("admin", response.getResolver());
         assertEquals(ReviewStatus.OPENED, response.getStatus());
     }
+    
+    @Test
+    public void updateReviewRequestTest() {
+        ReviewRequest response  = client.updateReviewRequest(updateRequest);
```
jdamerow: client is a mock, you're testing the mocking framework here.

jdamerow (changes requested):

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
kkrish39: This scenario will occur if a specific concept is requested for review for the second time.

In this case, there will be data in all the fields (resolving comment, name, resolver etc.). So Instead of nullifying each field, I deleted and added a new one.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
kkrish39: Sure. My bad. I'll do that.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
jdamerow: don't delete existing reviews. we want to keep a history of everything that has happened.

kkrish39: Okey. I was keeping track of only the last.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/core/ReviewRequest.java
kkrish39: Sure. Having this scenario, I should allow the resolver/requester only to comment without any resolve or reopen operation ?

For now, comment and reopen/resolve is tightly coupled.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ReviewRequestController.java
kkrish39: Comment will come in the POST body only. It will be pre populated.

Conceptpower+Spring/src/main/webapp/WEB-INF/views/home.jsp
kkrish39: Since we are adding exclamation icons dynamically after submitting request, the event handlers will not be created corresponding to it. ().click() will be separate for each corresponding html element.

Having $(document).on("click") there will be only one handler for the entire page matching the corresponding html element. It will handle the dynamically created elements as well.

Conceptpower+Spring/src/main/webapp/WEB-INF/views/home.jsp
kkrish39: Same reason as stated above.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/core/ReviewRequest.java
jdamerow: for now, just open/resolve requests with comments. later on, we'll probalby want to allow to just comment.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ReviewRequestController.java
jdamerow: right. in that case though, why don't you make it more generic and move the status to client site, so this could be used for any kind of status change?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
kkrish39: Using queryByExample will simply return a object but it won't point to the exact row value.

So updating that object and saving it back will create an another entry. Using predicate should avoid that. Correct me if I'm wrong.

jdamerow: actually, I might have been wrong. predicates might be using indexes and might be optimized. let's leave it like that, since we'll be moving to MySql anyways.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
```diff
@@ -30,8 +29,14 @@ public void init() {
...
     @Override
     public void store(ReviewRequest reviewRequest) {
-       
-        client.store(reviewRequest);
+        ReviewRequest storeRequest= new ReviewRequest();
+        
+        storeRequest.setConceptId(reviewRequest.getConceptId());
+        storeRequest.setStatus(reviewRequest.getStatus());
+        storeRequest.setRequest(reviewRequest.getRequest());
+        storeRequest.setRequester(reviewRequest.getRequester());
```
jdamerow: why are you doing this? If there is a good reason to create a new object here and simply copy the values into it, there needs to be a comment here explaining it.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
```diff
@@ -42,11 +47,74 @@ public ReviewRequest getReviewRequestForConcept(String conceptId){
         request.setConceptId(conceptId);
         
         List<ReviewRequest> reviewRequests = client.queryByExample(request);
-        
         if(reviewRequests!= null && reviewRequests.size()>0) {
-            return reviewRequests.get(0);
+            return reviewRequests.get(reviewRequests.size() - 1);
```
jdamerow: why are you taking the last instead of the first?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
```diff
@@ -42,11 +47,74 @@ public ReviewRequest getReviewRequestForConcept(String conceptId){
...
         }
+        
         return null;
     }
-   
+    
+    @Override
+    public List<ReviewRequest> getAllReviewsForaConcept(String conceptId){
```
jdamerow: just call this ```getAllReviews``` or ```getAllReviewsForConcept```

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
```diff
@@ -42,11 +47,74 @@ public ReviewRequest getReviewRequestForConcept(String conceptId){
...
+        ReviewRequest request = new ReviewRequest();
+        request.setConceptId(conceptId);
+        
+        List<ReviewRequest> reviewRequests = client.queryByExample(request);
+        
+        return reviewRequests;
```
jdamerow: merge with line 62

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
```diff
@@ -42,11 +47,74 @@ public ReviewRequest getReviewRequestForConcept(String conceptId){
...
+            private static final long serialVersionUID = 6495914730735826451L;
+
+            @Override
+            public boolean match(ReviewRequest review) {
+                return review.getConceptId().equals(reviewRequest.getConceptId());
+            }
+            
+        });
+       
+        List<String> comments  = requests.get(requests.size() - 1).getResolvingComment();
```
jdamerow: since you're not sorting by creation date or anything, I don't think you can make the assumption the last one is necessarily the latest one.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
```diff
@@ -42,11 +47,74 @@ public ReviewRequest getReviewRequestForConcept(String conceptId){
...
+        comments.addAll(reviewRequest.getResolvingComment());
+        
+        requests.get(requests.size() - 1).setResolvingComment(comments);
+        requests.get(requests.size() - 1).setStatus(reviewRequest.getStatus());
+        requests.get(requests.size() - 1).setResolver(reviewRequest.getResolver());
```
jdamerow: this is all business logic and needs to be done in manager. Ideally you do not need an extra db method for this. You update the request in the manager and simply call ```store``` on the db class.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
```diff
@@ -42,11 +47,74 @@ public ReviewRequest getReviewRequestForConcept(String conceptId){
...
+        
+        ReviewRequest updatedRequest = requests.get(requests.size() - 1);
+        
+        client.store(requests.toArray());
+        client.commit();
+        
+        return updatedRequest;
```
jdamerow: all my comments from above apply

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
jdamerow: not done

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/core/ReviewRequest.java
jdamerow: ```resolvingComment``` is not a good name for this property since it's not just for resolving comments, but potentially for any discussion in between as well.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ReviewRequestController.java
```diff
@@ -21,12 +22,33 @@
...
+        ReviewRequest updatedRequest = requestsMgr.reopenReview(reviewRequest);
+        
+        return updatedRequest;
+    }
+    
+    @RequestMapping(value = "/auth/request/getallreviews/{id}", method = RequestMethod.GET)
+    public @ResponseBody List<ReviewRequest> getAllReviews(@PathVariable String id, Principal principal){
+        List<ReviewRequest> listOfReviews = requestsMgr.getAllReviews(id);
+        
+        return listOfReviews;
```
jdamerow: merge with line 50

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ReviewRequestController.java
```diff
@@ -21,12 +22,33 @@
...
+    }
+    
+    @RequestMapping(value = "/auth/request/reopen", method = RequestMethod.POST )
+    public @ResponseBody ReviewRequest reopenRequest(@ModelAttribute(value="reviewRequest") ReviewRequest reviewRequest,Principal principal) {
+        ReviewRequest updatedRequest = requestsMgr.reopenReview(reviewRequest);
+        
+        return updatedRequest;
+    }
+    
+    @RequestMapping(value = "/auth/request/getallreviews/{id}", method = RequestMethod.GET)
```
jdamerow: make that just ```/auth/requests/{id}/all```

jdamerow: also, I would call ```id``` ```conceptId```, it is the concept id, right? id sounds like the request id.

Conceptpower+Spring/src/main/webapp/WEB-INF/views/home.jsp
```diff
@@ -225,20 +278,85 @@ $(document).ready(function(){
...
+				type:"POST",
+				url:"${pageContext.servletContext.contextPath}/auth/request/resolve?${_csrf.parameterName}=${_csrf.token}",
+				data: "resolvingComment=" + resolveComment + "&conceptId="+conceptId + "&status=RESOLVED",
+				success: function(response){
+					$('#requestModal').modal('hide');
+			    	$('#resolveComment').val('');
+					conceptId = conceptId.replace(/'/g, "\\'");
+			        var commentTd = $("#comment-" + conceptId);
+			        
+					commentTd.html('<div data-concept-id="'+conceptId+'" title="Add a review request" data-toggle="modal" data-target="#myModal" class="fa fa-comment" style="color:#19586B"></div>'
```
jdamerow: indentation

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/db/DBRequestClientTest.java
```diff
@@ -24,19 +28,66 @@
...
+    private ReviewRequest updateRequest = new ReviewRequest();
+
+    private ReviewRequest reopenRequest  = new ReviewRequest();
+    
     @Before
     public void init() {
         MockitoAnnotations.initMocks(this);
         
-        objectContainerClient = Mockito.mock(ObjectContainer.class);
+        client = Mockito.mock(DBRequestClient.class);
```
jdamerow: since DBRequestClient is the class being test, it cannot be mocked.

jdamerow (changes requested):

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
kkrish39: This I'm not clear. We need to update the already present data in the database. If we are updating the data in the manager and calling the store in here will add duplicate rows right ?

jdamerow: that is probably true for db4o but shouldn't for hibernate. So once we switch, the manager should only make an update call to the db class. For hibernate you might have to do the retrieve and copy, but that should be a general update method (not specific to this usecase).

kkrish39: Yeah. Then the current flow should work right. I have the updated request which I'm sending from Manager. Because of db4o, I'm retrieving the required data to be updated setting values to it and I'm calling store. In case of Hibernate, we should be able to add directly after include a unique id parameter

jdamerow: I think the issue, which I have with all of this is probably related more to the data model which bleeds into the storing procedure here. Each comment will have a creator and a creation date (which they don't have here, since they are just strings). Once the comments are proper objects, you can directly store them in the database and update modified requests.

kkrish39: Ah Yeah, I'm adding that as well. But even if I add those attributes I think we need to follow how we did before only in case of db4o. We need to somehow point to the place where the last ReviewRequest of given concept. Upon finding that we need to perform set() operation.

jdamerow: The user should see the last request (by creation date) and then you have an id for that request. There would be no list involved, just retrieve request by id and resolve it/add comment.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/RequestsManager.java
```diff
@@ -1,16 +1,23 @@
...
 import edu.asu.conceptpower.app.core.IRequestsManager;
 import edu.asu.conceptpower.app.db4o.IRequestsDBManager;
+import edu.asu.conceptpower.core.Comment;
 import edu.asu.conceptpower.core.ReviewRequest;
 
 @Service
 public  class RequestsManager implements IRequestsManager{
 
-  
+    protected final String LIST_PREFIX = "REVIEW";
```
jdamerow: why is this called LIST_PREFIX?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/RequestsManager.java
```diff
@@ -29,6 +36,83 @@ public void addReviewRequest(ReviewRequest newReviewRequest) {
      */
     public ReviewRequest getReview(String conceptId) {
         
-        return dbClient.getReviewRequestForConcept(conceptId);
+        ReviewRequest request = new ReviewRequest();
+        request.setConceptId(conceptId);
+        
+        //ArrayList is initialized to make the instance mutable for sorting
```
jdamerow: there is no list initialized? you get a list from the db client.

jdamerow: ahh, this comment needs to be moved down

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
```diff
@@ -29,24 +28,32 @@ public void init() {
...
     @Override
-    public void store(ReviewRequest reviewRequest) {
-       
+    public void store(ReviewRequest reviewRequest) {  
         client.store(reviewRequest);
         client.commit();
     }
 
+    
+    public List<ReviewRequest> getReview(ReviewRequest request) {
```
jdamerow: this method should take an id and return one object (the one with the given id).

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/RequestsManager.java
```diff
@@ -29,6 +36,83 @@ public void addReviewRequest(ReviewRequest newReviewRequest) {
...
+            return null;
+        }
+
+        List<Comment> comments  = responses.get(0).getComments();
+        comments.addAll(reviewRequest.getComments());
+        responses.get(0).setResolver(reviewRequest.getResolver());
+        responses.get(0).setComments(comments);
+        responses.get(0).setStatus(reviewRequest.getStatus());
+        
+        dbClient.updateReviewRequest(responses.toArray());
```
jdamerow: a method called "updateReviewRequest" (singular) should not take am array (plural)

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ReviewRequestController.java
```diff
@@ -21,12 +24,38 @@
...
         requestsMgr.addReviewRequest(reviewRequest);
         
         return reviewRequest;
     }
+    
+    @RequestMapping(value = "/auth/request/resolve", method = RequestMethod.POST )
+    public @ResponseBody ReviewRequest resolveRequest(@RequestBody ReviewRequest reviewRequest,Principal principal) {
+        reviewRequest.setResolver(principal.getName());
+        reviewRequest.getComments().get(0).setCreatedBy(principal.getName());
+        reviewRequest.getComments().get(0).setCreatedAt(LocalDateTime.now());
```
jdamerow: why are you setting these two properties at the first comment? Are you assuming that the first comment is a new one? ReviewRequest is the data send from the front end, right? So you assume that whatever is being send from the front end is exactly what's in the database plus a resolving comment? You cannot assume that. I could manipulate the data being sent (e.g. deleting all comments) and you would just store it in the database without checking. You can't trust what's coming from the frontend. Also, what if someone else resolved a request in the meantime? You'd loose their resolution comment and resolve it again.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ReviewRequestController.java
```diff
@@ -21,12 +24,38 @@
...
+        return updatedRequest;
+    }
+    
+    @RequestMapping(value = "/auth/request/reopen", method = RequestMethod.POST )
+    public @ResponseBody ReviewRequest reopenRequest(@RequestBody ReviewRequest reviewRequest,Principal principal) {
+        reviewRequest.getComments().get(0).setCreatedBy(principal.getName());
+        reviewRequest.getComments().get(0).setCreatedAt(LocalDateTime.now());
+        reviewRequest.setRequester(principal.getName());
+        
+        ReviewRequest updatedRequest = requestsMgr.updateReview(reviewRequest);
```
jdamerow: same as above

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ReviewRequestController.java
jdamerow: still here

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ReviewRequestController.java
```diff
@@ -21,12 +24,38 @@
...
+    
+    @RequestMapping(value = "/auth/request/resolve", method = RequestMethod.POST )
+    public @ResponseBody ReviewRequest resolveRequest(@RequestBody ReviewRequest reviewRequest,Principal principal) {
+        reviewRequest.setResolver(principal.getName());
+        reviewRequest.getComments().get(0).setCreatedBy(principal.getName());
+        reviewRequest.getComments().get(0).setCreatedAt(LocalDateTime.now());
+        
+        ReviewRequest updatedRequest = requestsMgr.updateReview(reviewRequest);
+        
+        return updatedRequest;
```
jdamerow: merge with line 40

jdamerow (changes requested):

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ReviewRequestController.java
kkrish39: Julia, In here I'm assuming that I'll be getting only one comment, since it's tightly coupled with Resolve/Reopen Request. And I'm only appending it to the already present comments for this specific reviewRequest.

jdamerow: but that will change when we add a commenting feature.

kkrish39: So, I should have a separate rest end point for adding comments and fire two calls(One to save comments and one for changing status) while resolving or reopening request for now ?

jdamerow: no. You assume that the review request you're getting from the front end is valid the way you receive it and that it doesn't have more than one comment, right? Two issues with that:
1) You are not validating what's being passed from the front end. I could set the status to whatever I like for example and you would simply store that in the database.
2) the same way, I could either add a bunch of comments to the review and you would store them all, or once we have a commenting feature, the passed review might originally have several comments on them and I could delete them all by sending modified data to the database.

kkrish39: Sure. I think now I got the point. I'll keep a check in the controller.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/IRequestsManager.java
```diff
@@ -20,6 +22,16 @@
...
     public abstract ReviewRequest getReview(String conceptId);
-
     
+    /**
+     * @param conceptId - the conceptId of the concept word
+     * @return List<ReviewRequest> - list of reviewRequests corresponding to a conceptId.
+     */
+    public List<ReviewRequest> getAllReviews(String conceptId);
+    /**
+     * Method to Resolve the reviewRequest
```
jdamerow: the method does not just resolve a review request, but update it.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/IRequestsManager.java
```diff
@@ -20,6 +22,16 @@
      * @return ReviewRequest - the ReviewRequest created for the concept with the provided conceptId.
      */
     public abstract ReviewRequest getReview(String conceptId);
-
     
+    /**
+     * @param conceptId - the conceptId of the concept word
```
jdamerow: Should be:
The id of the concept for which review requests are being retrieved.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/RequestsManager.java
```diff
@@ -29,6 +36,84 @@ public void addReviewRequest(ReviewRequest newReviewRequest) {
      */
     public ReviewRequest getReview(String conceptId) {
```
jdamerow: Since that now gets the latest review request, maybe rename the method.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/RequestsManager.java
```diff
@@ -29,6 +36,84 @@ public void addReviewRequest(ReviewRequest newReviewRequest) {
...
+    /* (non-Javadoc)
+     * @see edu.asu.conceptpower.app.core.IRequestsManager#updateReview(edu.asu.conceptpower.core.ReviewRequest)
+     */
+    
+    public ReviewRequest updateReview(ReviewRequest reviewRequest) {
+        ReviewRequest request = new ReviewRequest();
+        
+        request.setId(reviewRequest.getId());
+        
+        ReviewRequest responses = dbClient.getReview(request);
```
jdamerow: why is this called "responses" if it's just one? Also, why responses and not storedRequest or something?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/RequestsManager.java
```diff
@@ -29,6 +36,84 @@ public void addReviewRequest(ReviewRequest newReviewRequest) {
...
+        
+        if(responses == null) {
+            return null;
+        }
+
+        List<Comment> comments  = responses.getComments();
+        comments.addAll(reviewRequest.getComments());
+        responses.setResolver(reviewRequest.getResolver());
+        responses.setComments(comments);
```
jdamerow: not needed

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/RequestsManager.java
```diff
@@ -29,6 +36,84 @@ public void addReviewRequest(ReviewRequest newReviewRequest) {
...
+        request.setId(reviewRequest.getId());
+        
+        ReviewRequest responses = dbClient.getReview(request);
+        
+        if(responses == null) {
+            return null;
+        }
+
+        List<Comment> comments  = responses.getComments();
+        comments.addAll(reviewRequest.getComments());
```
jdamerow: merge with line 86

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
```diff
@@ -29,24 +28,38 @@ public void init() {
...
 
-    @Override
-    public ReviewRequest getReviewRequestForConcept(String conceptId){
-        
-        ReviewRequest request = new ReviewRequest();
-        request.setConceptId(conceptId);
-        
-        List<ReviewRequest> reviewRequests = client.queryByExample(request);
+    
+    public ReviewRequest getReview(ReviewRequest request) {
```
jdamerow: why do you pass a reviewRequest object instead of the id?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
```diff
@@ -29,24 +28,38 @@ public void init() {
...
+        if(responses == null || responses.size() == 0) {
+            return null;
         }
-        return null;
+        
+        return responses.get(0);
+    }
+    
+    @Override
+    public List<ReviewRequest> getAllReviews(ReviewRequest reviewRequest){
```
jdamerow: this method should take a concept id and then build the example object here, otherwise you'll have to change the logic of the manager class again when the backend is exchanged.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
```diff
@@ -29,24 +28,38 @@ public void init() {
...
+        return client.queryByExample(reviewRequest);
+    }
+    
+    @Override
+    public void updateReviewRequest(Object reviewRequest) {
```
jdamerow: why is that method taking an object and not a reviewrequest?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/core/Comment.java
```diff
@@ -0,0 +1,33 @@
+package edu.asu.conceptpower.core;
+
+import java.time.LocalDateTime;
+
+public class Comment{
+	private String comment;
+	private String createdBy;
+	private LocalDateTime createdAt;
```
jdamerow: LocalDateTime doesn't have any timezone information attached to it, which doesn't work in most applications (including this). Use OffsetDateTime instead.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/core/ReviewRequest.java
jdamerow: getters and setters should have the same name as the variables they belong to.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ReviewRequestController.java
```diff
@@ -21,12 +27,39 @@
...
-        
+        reviewRequest.setCreatedAt(LocalDateTime.now());
         requestsMgr.addReviewRequest(reviewRequest);
         
         return reviewRequest;
     }
+    
+    @RequestMapping(value = "/auth/request/resolve", method = RequestMethod.POST )
+    public @ResponseBody ResponseEntity<ReviewRequest> resolveRequest(@RequestBody ReviewRequest reviewRequest,Principal principal) {
+        if(reviewRequest.getStatus() != ReviewStatus.RESOLVED || reviewRequest.getComments().size() != 1) {
```
jdamerow: as soon as we have functionality that allows to just add a comment to a request, this method will no longer work.
This method should take a review request id, status, and comment and if the review request for the id is not yet resolved, resolve it (set resolver, resolution time, add the comment, set the status).

Conceptpower+Spring/src/main/webapp/WEB-INF/views/home.jsp
```diff
@@ -191,54 +191,192 @@ function prepareMergeConcept(conceptId) {
...
+        	}
+        }
+    });
+}
+
 $(document).ready(function(){
-    $(".fa-exclamation-triangle").click(function() {
-		$("#fetchRequests").val($(this).data("request"));
+	
+    var reviewId;
```
jdamerow: global variables are always a source for errors; set data attributes on the elements you're working with and use those instead.

jdamerow (changes requested):

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/RequestsManager.java
kkrish39: Julia, We need to append the latest comment to the already present comments list right ?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/web/ReviewRequestController.java
kkrish39: Shouldn't we have a separate endpoint when it comes to adding only comment ? 

Correct me if I'm wrong. In the above case, we are assuming that the first comment is tightly coupled with Resolving/Reopening the review request and the later coming comments are added without any status change.

But what if the user just need to update the comment in the beginning without any status change ?

jdamerow: Yes, there will be a separate endpoint for adding comments. However, the reviewRequest is the object that can be commented on before it's being resolved, so 
1. a user creates a review request (with a note why it needs to be reviewed)
2. another user might make a comment about how it should be changed
3. the first user might make anotehr comment in response
4. step 3 and 4 might repeat or another user might make a comment
5. eventually a user will resolve the request; the request will have multiple comments by then.

jdamerow: editing a comment will also be a different endpoint if we allow that.

kkrish39: In Step 5: There is the possibility of having 0 or 1 comment only right in the request while resolving ? Because rest of the comments will already get added by separate end points. We must be sure that the reviewRequest coming to this method should not be more that 1.

One more clarification, if I'm getting a status other that "RESOLVED", you want me to update with the correct status instead of sending 400 ?

jdamerow: Simply don't use the review request object here. It doesn't make sense to use incomplete review request objects, when all you need is the comment and the review request id. It's just confusing. If you want to use an object, create a dedicated class for the information, otherwise you can get the id from the url and the comment as request parameter. Since your endpoint is called /resolve you don't even need a status.

kkrish39:
While using OffsetDateTime, occasionally the date is getting reseted to Zero and It couldn't generate a valid JSON. So it throws the following error. I debugged and tried multiple possibilities and I clearly couldn't find the root cause. LocalDateTime is working fine.

"org.springframework.http.converter.HttpMessageNotWritableException: Could not write JSON: Invalid value for MonthOfYear: 0; nested exception is com.fasterxml.jackson.databind.JsonMappingException: Invalid value for MonthOfYear: 0"

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/RequestsManager.java
```diff
@@ -20,15 +29,102 @@
...
+        request.setConceptId(conceptId);
+        
+        List<ReviewRequest> reviewRequests = dbClient.getAllReviews(request);
+        
+        if(reviewRequests!= null && reviewRequests.size() > 0) {
+            reviewRequests = new ArrayList<>(reviewRequests);
+            Collections.sort(reviewRequests, (x, y) -> x.getCreatedAt().compareTo(y.getCreatedAt()));
+        }
+        
+        return reviewRequests;
```
jdamerow: this seems to be almost the same code as "getLatestReview" -> DRY

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/RequestsManager.java
```diff
@@ -20,15 +29,102 @@
...
+        
+        ReviewRequest storedRequest = dbClient.getReview(request);
+        
+        if(storedRequest == null) {
+            return null;
+        }
         
-        return dbClient.getReviewRequestForConcept(conceptId);
+        if(comments != null && comments.size() > 0) {
+            //There is only a possibility of single comment at a time
```
jdamerow: if that's the case, why is a list being passed?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
jdamerow: ?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
jdamerow: ?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/core/ReviewRequest.java
jdamerow: ?

jdamerow (changes requested):

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/RequestsManager.java
kkrish39: Since I was getting in the form of ReviewRequest Object before, I had this. Removed and pushed now.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
jdamerow: ?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/DBRequestClient.java
jdamerow: ?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/core/ReviewRequest.java
jdamerow: ?

jdamerow (changes requested):
not all comments addressed

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/core/impl/RequestsManagerTest.java
```diff
@@ -19,22 +27,43 @@
...
+    
+    private ReviewRequest request = new ReviewRequest();
+    
     @Before
     public void setup() {
         MockitoAnnotations.initMocks(this);
         
+        List<ReviewRequest> responseList = new ArrayList<>();
+        ReviewRequest updatedRequest = new ReviewRequest();
+        request.setId("REVIEW1234-thisissupposedtobesecret");
```
jdamerow: order of lines is very confusing. keep the things together that belong together: updated request with the rest of the updatedrequest lines, response list initialization below, where it's needed, request separate from that.

jdamerow: also, id's are not secret, so no reason to call it that.

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/core/impl/RequestsManagerTest.java
```diff
@@ -19,22 +27,43 @@
...
     public void test_addReviewRequest_success() {
         
         ReviewRequest reviewRequest = new ReviewRequest();
         reviewRequest.setStatus(ReviewStatus.OPENED);
+        reviewRequest.setId("REVIEW1234-thisissupposedtobesecret");
         reviewRequest.setRequester("admin");
         reviewRequest.setConceptId("WID-01006675-N-02-mental_test");
         reviewRequest.setRequest("Testing Method");
            
+        Mockito.when(dbClient.getReview("REVIEW1234-thisissupposedtobesecret")).thenReturn(null);
```
jdamerow: why is this here? the default value is null in mockito if you haven't defined it otherwise.

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/core/impl/RequestsManagerTest.java
```diff
@@ -49,9 +78,11 @@ public void test_getReview_success() {
         reviewRequest.setConceptId("WID-10126926-N-05-Einstein");
         reviewRequest.setRequest("Testing Method2");
         
-        Mockito.when(dbClient.getReviewRequestForConcept("WID-10126926-N-05-Einstein")).thenReturn(reviewRequest);
+        List<ReviewRequest> response = new ArrayList<>();
+        response.add(reviewRequest);
+        Mockito.when(dbClient.getAllReviews("WID-10126926-N-05-Einstein")).thenReturn(response);
```
jdamerow: is this necessary? you already set up returning a response list in the before method, dont' you?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/RequestsManager.java
```diff
@@ -20,15 +29,84 @@
      * @see edu.asu.conceptpower.app.core.IRequestsManager#addReviewRequest(edu.asu.conceptpower.core.ReviewRequest)
      */
     public void addReviewRequest(ReviewRequest newReviewRequest) {
-        
+        newReviewRequest.setId(generateId(REVIEW_PREFIX));
         dbClient.store(newReviewRequest);
     }
     
     /* (non-Javadoc)
      * @see edu.asu.conceptpower.app.core.IRequestsManager#getReview(java.lang.String)
```
jdamerow: needs to be adjusted

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/core/impl/RequestsManagerTest.java
```diff
@@ -49,9 +78,11 @@ public void test_getReview_success() {
...
 
-        ReviewRequest fetchedReview = requestManager.getReview("WID-10126926-N-05-Einstein");
+        ReviewRequest fetchedReview = requestManager.getLatestReview("WID-10126926-N-05-Einstein");
```
jdamerow: test method needs to be renamed according to tested method

jdamerow: also, the important thing here is that getLatestReview gives you the most recent review. You are not testing that.

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/core/impl/RequestsManagerTest.java
```diff
@@ -61,17 +92,51 @@ public void test_getReview_success() {
...
-        reviewRequest.setStatus(ReviewStatus.OPENED); 
-        reviewRequest.setRequester("admin");
-        reviewRequest.setConceptId("WID-10126926-N-05-Einstein");
-        reviewRequest.setRequest("Testing Method2");
-        
-        Mockito.when(dbClient.getReviewRequestForConcept("WID-10126926-N-05-Einstein")).thenReturn(reviewRequest);
+        Mockito.when(dbClient.getAllReviews("WID-10126926-N-05-Thomas")).thenReturn(null);
 
-        ReviewRequest fetchedReview = requestManager.getReview("WID-10126926-N-05-Thomas");
+        ReviewRequest fetchedReview = requestManager.getLatestReview("WID-10126926-N-05-Thomas");
```
jdamerow, line 96: rename test method

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/core/impl/RequestsManagerTest.java
```diff
@@ -61,17 +92,51 @@ public void test_getReview_success() {
...
         
         Assert.assertNull(fetchedReview);
     }
+    
+    @Test
+    public void updateReviewTest() {
```
jdamerow: not how we name testing method, please read our guidelines

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/core/impl/RequestsManagerTest.java
```diff
@@ -61,17 +92,51 @@ public void test_getReview_success() {
...
+        comment.setComment("Sample Comment");
+        
+        ReviewRequest response  = requestManager.updateReview("REVIEW1234-thisissupposedtobesecret2", ReviewStatus.OPENED, comment, OffsetDateTime.now(), "admin");
+        
+        assertNotNull(response);
+        
+        assertEquals("WID-10126926-N-05-Einstein", response.getConceptId());
+        assertEquals("REVIEW1234-thisissupposedtobesecret2", response.getId());
+        assertEquals("Review the Spelling again", response.getRequest());
+        assertEquals("admin", response.getRequester());
```
jdamerow, line 114: the important thing here is that the comment is added to the review, isn't it? You're not testing that.

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/core/impl/RequestsManagerTest.java
```diff
@@ -61,17 +92,51 @@ public void test_getReview_success() {
...
+    
+    @Test
+    public void getAllReviewsTest() {
+        List<ReviewRequest> response = requestManager.getAllReviews("WID-10126926-N-05-Einstein");
+        
+        assertNotNull(response);
+        
+        assertEquals(1, response.size());
+        assertEquals("Review the Spelling again", response.get(0).getRequest());
+        assertEquals("REVIEW1234-thisissupposedtobesecret2", response.get(0).getId());
```
jdamerow: since you're using the ids so much all over the place, they should be constants.

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/core/impl/RequestsManagerTest.java
```diff
@@ -61,17 +92,51 @@ public void test_getReview_success() {
...
+    }
+    
+    @Test
+    public void getReviewTest() {
```
jdamerow: what is this testing? what is different from test_getReview_success?

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/db/DBRequestClientTest.java
```diff
@@ -1,53 +1,42 @@
...
-        reviewRequest.setResolver("admin");
-        reviewRequest.setStatus(ReviewStatus.OPENED);
-
-        Mockito.when(client.getReviewRequestForConcept(reviewRequest.getConceptId())).thenReturn(reviewRequest);
     }
+    
+    @Test
+    public void getAllReviewsTest() {
+        client.getAllReviews("WID-10126926-N-05-Einstein");
+        Mockito.verify(objectContainerClient).queryByExample(ArgumentMatchers.any(ReviewRequest.class));
```
jdamerow: you want to make sure that queryByExample is called with an example that has the correct id, otherwise your test will succeed even if the queried is is wrong.

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/db/DBRequestClientTest.java
```diff
@@ -1,53 +1,42 @@
...
-        assertNotNull(response);
-        
-        assertEquals("WID-10126926-N-05-Einstein", response.getConceptId());
-        assertEquals("Review the Spelling", response.getRequest());
-        assertEquals("user2", response.getRequester());
-        assertEquals("admin", response.getResolver());
-        assertEquals(ReviewStatus.OPENED, response.getStatus());
+    public void getReviewTest() {
+        client.getReview("REVIEW1234-thisissupposedtobesecret2");
+        Mockito.verify(objectContainerClient).query(Mockito.<Predicate<ReviewRequest>>anyObject());
```
jdamerow: same as above

jdamerow (changes requested):

jdamerow:
Jenkins, test this please.

jdamerow:
Jenkins, test this please.

jdamerow:
Search throws a null pointer.
```
ERROR: edu.asu.conceptpower.web.util.ExceptionHandler - ExceptionHandler caught exception.
java.lang.NullPointerException
	at edu.asu.conceptpower.app.core.impl.RequestsManager.lambda$getAllReviews$2(RequestsManager.java:57)
	at java.util.TimSort.countRunAndMakeAscending(TimSort.java:355)
	at java.util.TimSort.sort(TimSort.java:220)
	at java.util.Arrays.sort(Arrays.java:1512)
	at java.util.ArrayList.sort(ArrayList.java:1462)
	at java.util.Collections.sort(Collections.java:175)
	at edu.asu.conceptpower.app.core.impl.RequestsManager.getAllReviews(RequestsManager.java:57)
	at edu.asu.conceptpower.app.core.impl.RequestsManager.getLatestReview(RequestsManager.java:40)
	at edu.asu.conceptpower.app.wrapper.impl.ConceptEntryWrapperCreator.createWrappers(ConceptEntryWrapperCreator.java:89)
	at edu.asu.conceptpower.web.ConceptSearchController.search(ConceptSearchController.java:121)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.springframework.web.method.support.InvocableHandlerMethod.doInvoke(InvocableHandlerMethod.java:190)
	at org.springframework.web.method.support.InvocableHandlerMethod.invokeForRequest(InvocableHandlerMethod.java:138)
	at org.springframework.web.servlet.mvc.method.annotation.ServletInvocableHandlerMethod.invokeAndHandle(ServletInvocableHandlerMethod.java:106)
	at org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerAdapter.invokeHandlerMethod(RequestMappingHandlerAdapter.java:888)
	at org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerAdapter.handleInternal(RequestMappingHandlerAdapter.java:793)
	at org.springframework.web.servlet.mvc.method.AbstractHandlerMethodAdapter.handle(AbstractHandlerMethodAdapter.java:87)
	at org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:1040)
	at org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:943)
	at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:1006)
	at org.springframework.web.servlet.FrameworkServlet.doGet(FrameworkServlet.java:898)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:635)
	at org.springframework.web.servlet.FrameworkServlet.service(FrameworkServlet.java:883)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:742)
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:231)
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
	at org.apache.tomcat.websocket.server.WsFilter.doFilter(WsFilter.java:52)
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:193)
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
	at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:320)
	at org.springframework.security.web.access.intercept.FilterSecurityInterceptor.invoke(FilterSecurityInterceptor.java:126)
	at org.springframework.security.web.access.intercept.FilterSecurityInterceptor.doFilter(FilterSecurityInterceptor.java:90)
	at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:334)
	at org.springframework.security.web.access.ExceptionTranslationFilter.doFilter(ExceptionTranslationFilter.java:118)
	at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:334)
	at org.springframework.security.web.session.SessionManagementFilter.doFilter(SessionManagementFilter.java:137)
	at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:334)
	at org.springframework.security.web.authentication.AnonymousAuthenticationFilter.doFilter(AnonymousAuthenticationFilter.java:111)
	at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:334)
	at org.springframework.security.web.servletapi.SecurityContextHolderAwareRequestFilter.doFilter(SecurityContextHolderAwareRequestFilter.java:158)
	at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:334)
	at org.springframework.security.web.savedrequest.RequestCacheAwareFilter.doFilter(RequestCacheAwareFilter.java:63)
	at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:334)
	at org.springframework.security.web.authentication.www.BasicAuthenticationFilter.doFilterInternal(BasicAuthenticationFilter.java:154)
	at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:119)
	at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:334)
	at org.springframework.security.web.authentication.AbstractAuthenticationProcessingFilter.doFilter(AbstractAuthenticationProcessingFilter.java:200)
	at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:334)
	at org.springframework.security.web.authentication.logout.LogoutFilter.doFilter(LogoutFilter.java:116)
	at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:334)
	at org.springframework.security.web.csrf.CsrfFilter.doFilterInternal(CsrfFilter.java:117)
	at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:119)
	at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:334)
	at org.springframework.security.web.header.HeaderWriterFilter.doHeadersAfter(HeaderWriterFilter.java:92)
	at org.springframework.security.web.header.HeaderWriterFilter.doFilterInternal(HeaderWriterFilter.java:77)
	at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:119)
	at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:334)
	at org.springframework.security.web.context.request.async.WebAsyncManagerIntegrationFilter.doFilterInternal(WebAsyncManagerIntegrationFilter.java:56)
	at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:119)
	at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:334)
	at org.springframework.security.web.context.SecurityContextPersistenceFilter.doFilter(SecurityContextPersistenceFilter.java:105)
	at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:334)
	at org.springframework.security.web.FilterChainProxy.doFilterInternal(FilterChainProxy.java:215)
	at org.springframework.security.web.FilterChainProxy.doFilter(FilterChainProxy.java:178)
	at org.springframework.web.filter.DelegatingFilterProxy.invokeDelegate(DelegatingFilterProxy.java:358)
	at org.springframework.web.filter.DelegatingFilterProxy.doFilter(DelegatingFilterProxy.java:271)
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:193)
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:166)
```

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/core/impl/RequestsManagerTest.java
jdamerow: ?

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/core/impl/RequestsManagerTest.java
jdamerow: ?

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/core/impl/RequestsManagerTest.java
jdamerow: ?

jdamerow (changes requested):
You don't seem to have addressed my previous comments.

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/db/DBRequestClientTest.java
kkrish39: Since the object instance is created inside the method, we can't check with an equivalent class object. I tried lot of way and ended up with this. Please let me know if we can do it in any other way.

jdamerow: https://www.baeldung.com/mockito-argument-matchers section 4 describes your problem I believe.

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/core/impl/RequestsManagerTest.java
```diff
@@ -19,59 +27,117 @@
...
+        assertEquals("Sample Comment",response.getComments().get(0).getComment());
+        assertEquals(CONCEPT_ID1, response.getConceptId());
+        assertEquals(REVIEW_ID1, response.getId());
+        assertEquals("Review the Spelling again", response.getRequest());
+        assertEquals("admin", response.getRequester());
+        assertEquals(ReviewStatus.OPENED, response.getStatus());
+    }
+    
+    @Test
+    public void test_updateReview_failure() {
```
jdamerow: updating a review can fail for several reasons, you should test each. and the method name should reflect the failure reason.

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/db/CustomMatchers/PredicateReviewRequestMatcher.java
```diff
@@ -0,0 +1,35 @@
+package edu.asu.conceptpower.app.db.CustomMatchers;
```
jdamerow: packages names are not uppercased.

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/db/CustomMatchers/PredicateReviewRequestMatcher.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * This class is a custom argument matcher to compare Predicate<ReviewRequest> classes
+ * 
+ * @author keerthivasan
+ *
+ */
+public class PredicateReviewRequestMatcher implements ArgumentMatcher<Predicate<ReviewRequest>> {
+ 
+    private Predicate<ReviewRequest> left;
```
jdamerow: bad variable name, what does left mean in this context?

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/db/DBRequestClientTest.java
```diff
@@ -1,53 +1,58 @@
...
-        
-        reviewRequest.setConceptId("WID-10126926-N-05-Einstein");
-        reviewRequest.setRequest("Review the Spelling");
-        reviewRequest.setRequester("user2");
-        reviewRequest.setResolver("admin");
-        reviewRequest.setStatus(ReviewStatus.OPENED);
 
-        Mockito.when(client.getReviewRequestForConcept(reviewRequest.getConceptId())).thenReturn(reviewRequest);
+        client.getReview(REVIEW_ID);
+        Mockito.verify(objectContainerClient).query(Mockito.argThat(new PredicateReviewRequestMatcher(request, REVIEW_ID)));
```
jdamerow, line 72: why do you verify that the objectContainerClient is called instead of making sure the returned review request object is correct?

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/db/DBRequestClientTest.java
```diff
@@ -1,53 +1,58 @@
...
+        ReviewRequest request = new ReviewRequest();
+        request.setConceptId(CONCEPT_ID);
         
-        assertEquals("WID-10126926-N-05-Einstein", response.getConceptId());
-        assertEquals("Review the Spelling", response.getRequest());
-        assertEquals("user2", response.getRequester());
-        assertEquals("admin", response.getResolver());
-        assertEquals(ReviewStatus.OPENED, response.getStatus());
+        client.getAllReviews(CONCEPT_ID);
+        Mockito.verify(objectContainerClient).queryByExample(Mockito.argThat(new ReviewRequestMatcher(request)));
```
jdamerow, line 85: same as above, you don't test what the method actually returns.

jdamerow (changes requested):

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/db/CustomMatchers/PredicateReviewRequestMatcher.java
kkrish39: I named the two objects to be compared as left and right. Left is the member variable and right will be used inside the matches() method.

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/db/DBRequestClientTest.java
kkrish39: Both query() and queryByExample() are expecting the return object in ObjectSet<Object> type. I couldn't type cast a List of reviewRequests or plain objects only. Thats why I only verified whether the specific function is called with the expected parameter.

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/db/CustomMatchers/PredicateReviewRequestMatcher.java
jdamerow: give them more telling names that left/right

Conceptpower+Spring/src/test/java/edu/asu/conceptpower/app/db/DBRequestClientTest.java
jdamerow: since objectset is an interface, why don't you let mockito mock it for you or simply provide a dummy implementation?

jdamerow:
test this please

jdamerow:
Searching for "einstein" still gives this error:
```
ERROR: edu.asu.conceptpower.web.util.ExceptionHandler - ExceptionHandler caught exception.
java.lang.NullPointerException
	at edu.asu.conceptpower.app.core.impl.RequestsManager.lambda$getAllReviews$2(RequestsManager.java:57)
	at java.util.TimSort.countRunAndMakeAscending(TimSort.java:355)
	at java.util.TimSort.sort(TimSort.java:220)
	at java.util.Arrays.sort(Arrays.java:1512)
	at java.util.ArrayList.sort(ArrayList.java:1462)
	at java.util.Collections.sort(Collections.java:175)
```

Also, we'll probably have to make some changes to the UI, but let's talk about that once it's in thymeleaf.

kkrish39:
A gentle remainder that this branch should be pushed to dev after pushing thyme leaf changes.

jdamerow:
test this please

jdamerow:
resolving a request fails, the logs say:
`WARN : org.springframework.web.servlet.PageNotFound - No mapping for POST /conceptpower-review/auth/request/resolve/
`

jdamerow:
- version only seems to show up on the home page not on the other pages
- I an add a synonym several times (click the Add button several times) and for some reason the font of the first synonym is different (smaller?) than of the others:
image
- copyright notice in right bottom corner should say 2020

jdamerow:
Jenkins, test this please.

jdamerow:
To be reverted back when thymeleaf changes are merged.

jdamerow:
Please merge develop and reopen.

jdamerow:
Make it so, Jenkins.

jdamerow:
there don't seem to be any frontend changes?

kkrish39:
> there don't seem to be any frontend changes?

Yes. We decided to work on the UI later on after we stabilize the Thymeleaf changes. I have not developed the Thymeleaf version of it yet. 

Do you have any suggestions on how the UI should look like? I remember, you saying that you have some comments on the previous UI. Please let me know. I can start working on this.

jdamerow:
This branch is only for CCP-253, right? The other tickets around the review comments are in separate branches? Or will this just contain all changes?

For CCP-253, just another entry in the menu that shows a searchable list of all concepts with open review comments. For the other tickets about creating, resolving, etc. there should be another column in the search interface after the user logged in, that has a button/link icon to add a new review comment (if there is none yet) that when clicked opens a modal to enter the review comment. If there is already a review comment, it should show an icon indicating that in the last column. Clicking on that icon should again show a modal with the review comment. The option to add a response comment or to resolve the comment.

kkrish39:
> This branch is only for CCP-253, right? The other tickets around the review comments are in separate branches? Or will this just contain all changes?
> 
> For CCP-253, just another entry in the menu that shows a searchable list of all concepts with open review comments. For the other tickets about creating, resolving, etc. there should be another column in the search interface after the user logged in, that has a button/link icon to add a new review comment (if there is none yet) that when clicked opens a modal to enter the review comment. If there is already a review comment, it should show an icon indicating that in the last column. Clicking on that icon should again show a modal with the review comment. The option to add a response comment or to resolve the comment.

It does have all the changes related to Review Request. Sure, I will work on that.

PR #207, opened 2020-09-01
Migration from db4o to SQL

kkrish39:
The backend logic to enable migration is completed. But from UI, we can only trigger to migrate ConceptEntry table. Working on it to add additional enhancements.


Conceptpower+Spring/pom.xml
```diff
@@ -38,16 +38,17 @@
 		<concept.list.url>https://www.dropbox.com/s/r82nymnt2247zz4/conceptLists.db?raw=1</concept.list.url>
 		<concept.types.url>https://www.dropbox.com/s/nwn15s84fzz37s5/conceptTypes.db?raw=1</concept.types.url>
 		<users.url>https://www.dropbox.com/s/p6h3974hmlouq0i/users.db?raw=1</users.url>
+		
+		
+				<!-- DB config -->
```
radhikanyati94, line 43: Indentation is missing. Remove extra empty lines.

Conceptpower+Spring/pom.xml
```diff
@@ -64,6 +65,15 @@
 				<admin.pw>${admin.password}</admin.pw>
 				<wordnet.install.path>${wordnet.path}</wordnet.install.path>
 				<lucene.path>${lucene.index}</lucene.path>
+				
+				
+				<!-- DB config -->
```
radhikanyati94, line 70: Indentation is missing.

Conceptpower+Spring/pom.xml
```diff
@@ -388,26 +398,49 @@
...
+		<!-- Spring Data JPA -->
+		<!-- https://mvnrepository.com/artifact/org.springframework.data/spring-data-jpa -->
+		<dependency>
+		    <groupId>org.springframework.data</groupId>
+		    <artifactId>spring-data-jpa</artifactId>
+		    <version>2.2.6.RELEASE</version>
+		</dependency>
+		
+		 <!-- DataSource (HikariCP) -->
+        <dependency>
```
radhikanyati94, line 423: Indentation is missing.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/RequestsManager.java
```diff
@@ -19,16 +29,85 @@
...
-    public void addReviewRequest(ReviewRequest newReviewRequest) {
+    public void addReviewRequest(ReviewRequest reviewRequest) {
+        reviewRequest.setId(generateId(REVIEW_PREFIX));
+        dbClient.store(reviewRequest);
+    }
+    
+    /* (non-Javadoc)
+     * @see edu.asu.conceptpower.app.core.IRequestsManager#getLatestReview(java.lang.String)
+     */
+    public ReviewRequest getLatestReview(String conceptId) {
```
radhikanyati94, line 40: Can conceptId be null or empty in any case?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/RequestsManager.java
```diff
@@ -19,16 +29,85 @@
...
+        return null;
     }
     
     /* (non-Javadoc)
-     * @see edu.asu.conceptpower.app.core.IRequestsManager#getReview(java.lang.String)
+     * @see edu.asu.conceptpower.app.core.IRequestsManager#getAllReviews(java.lang.String)
      */
-    public ReviewRequest getReview(String conceptId) {
+    public List<ReviewRequest> getAllReviews(String conceptId){
+        List<ReviewRequest> reviewRequests = dbClient.getAllReviews(conceptId);
```
radhikanyati94, line 54: Can conceptId be null or empty in any case?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/RequestsManager.java
```diff
@@ -19,16 +29,85 @@
...
+        dbClient.store(reviewRequest);
+    }
+    
+    /* (non-Javadoc)
+     * @see edu.asu.conceptpower.app.core.IRequestsManager#getLatestReview(java.lang.String)
+     */
+    public ReviewRequest getLatestReview(String conceptId) {
+        List<ReviewRequest> reviewRequests = getAllReviews(conceptId);
+       
+        if(reviewRequests!= null && reviewRequests.size() > 0) {
```
radhikanyati94, line 43: Is "reviewRequests.size() > 0" check required here? getAllReviews either returns list with more than 0 elements  or it will return null.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/RequestsManager.java
```diff
@@ -19,16 +29,85 @@
...
+        
+        return null;
+    }
+    
+    /* (non-Javadoc)
+     * @see edu.asu.conceptpower.app.core.IRequestsManager#updateReview(edu.asu.conceptpower.core.ReviewRequest)
+     */
+    
+    public ReviewRequest updateReview(String reviewId, ReviewStatus reviewStatus, Comment comment, OffsetDateTime createdAt,String updatedBy) {  
+        ReviewRequest storedRequest = dbClient.getReview(reviewId);
```
radhikanyati94, line 70: Can reviewId be null?

jdamerow (changes requested):
Let's use `PagingAndSortingRepository` for all the repositories. That will cut down significantly on the coding work (e.g. https://www.baeldung.com/spring-data-jpa-pagination-sorting). And let's also use the spring release train as used for example here: https://github.com/diging/citesphere/blob/story/CITE-103/citesphere/pom.xml That will make the version management of the plugins easier.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/RequestsManager.java
kkrish39: Good question. For all the null cases you've asked, there is a possibility because it's coming from the frontend. The database client handles that. We are checking the database response and making necessary changes.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/RequestsManager.java
kkrish39: We may not be sure. Just to keep the code defensive, it's good to check for null and empty cases. Instead of `reviewRequests.size() > 0` we can actually use inbuilt method (`reviewRequests.isEmpty()`) to check for empty case.

Conceptpower+Spring/pom.xml
kkrish39: Good catch! Done

PR #208, opened 2020-09-03
Migration to SQL

kkrish39:


Conceptpower+Spring/pom.xml
```diff
@@ -38,16 +39,15 @@
 		<concept.list.url>https://www.dropbox.com/s/r82nymnt2247zz4/conceptLists.db?raw=1</concept.list.url>
 		<concept.types.url>https://www.dropbox.com/s/nwn15s84fzz37s5/conceptTypes.db?raw=1</concept.types.url>
 		<users.url>https://www.dropbox.com/s/p6h3974hmlouq0i/users.db?raw=1</users.url>
+		<!-- DB config -->
+        <db.driver>com.mysql.jdbc.Driver</db.driver>
+        <db.database.url></db.database.url>
+        <db.user></db.user>
+        <db.password></db.password>
+        <hibernate.dialect>org.hibernate.dialect.MySQL5InnoDBDialect</hibernate.dialect>
+        <hibernate.log.sql>false</hibernate.log.sql>
```
jdamerow, line 48: let's reformat the whole file to even out the indentations

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/model/impl/ChangeEvent.java
```diff
@@ -0,0 +1,81 @@
...
+import javax.persistence.Table;
+
+import edu.asu.conceptpower.servlet.core.ChangeEvent.ChangeEventTypes;
+
+
+@Entity
+@Table(name="change_event")
+public class ChangeEvent implements Serializable {
+
+    private static final long serialVersionUID = 1635835636441631877L;
```
jdamerow, line 22: let's start at 1L.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/model/impl/Comment.java
```diff
@@ -0,0 +1,76 @@
...
+import javax.persistence.Id;
+import javax.persistence.JoinColumn;
+import javax.persistence.ManyToOne;
+import javax.persistence.Table;
+
+@Entity
+@Table(name="comment")
+public class Comment implements Serializable{
+    
+    private static final long serialVersionUID = -6643011518760788400L;
```
jdamerow, line 17: same as above and for all other serialVersionUIDs

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db4o/IRequestsDBManager.java
```diff
@@ -1,13 +1,23 @@
 package edu.asu.conceptpower.app.db4o;
 
+import java.util.List;
 
 import edu.asu.conceptpower.core.ReviewRequest;
 
+
+/**
+ * 
+ * @author Keerthivasan
```
jdamerow, line 10: needs full name and description

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/migration/impl/MigrateToSql.java
```diff
@@ -0,0 +1,144 @@
...
+        
+        return count;
+    }
+    
+    @Async
+    public int migrateConceptListTable() {
+        List<ConceptList> conceptListDump = listManager.getAllConceptLists();
+        int count = 0;
+        
+        if(conceptListDump != null && !conceptListDump.isEmpty()) {
```
jdamerow, line 116: even if `conceptListDump` is empty, you can still iterate over it.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/migration/impl/ModelMapperUtil.java
```diff
@@ -0,0 +1,131 @@
...
+ * @author Keerthivasan
+ * 
+ */
+
+public class ModelMapperUtil {
+    
+    /*Avoid Instantiation of this utility class*/
+    private ModelMapperUtil() {}
+    
+    public static List<edu.asu.conceptpower.app.core.model.impl.ConceptType> conceptTypeMapper(ConceptType[] conceptTypeList){
```
jdamerow, line 24: we're using spring, so no static methods. make this class a component or service.

jdamerow: also, I would not pass an array. you would end up holding two potentially very long arrays/lists in memory instead of doing it one by one. (meaning you would probably still have one big array, but you could write them one by one into the database to avoid having twice the amount of objects in memory).

jdamerow: and last, `conceptTypeMapper` is a bad method name. Method names are usually verbs.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/migration/impl/MigrateToSql.java
```diff
@@ -0,0 +1,144 @@
...
+    
+    @Async
+    public int migrateConceptTypeTable() {
+        ConceptType[] conceptTypeDump = typesManager.getAllTypes();
+        int count = 0;
+        
+        if(conceptTypeDump != null && conceptTypeDump.length > 0) {
+            List<edu.asu.conceptpower.app.core.model.impl.ConceptType> mappedConceptType = ModelMapperUtil.conceptTypeMapper(conceptTypeDump);
+            
+            for(edu.asu.conceptpower.app.core.model.impl.ConceptType c : mappedConceptType) {
```
jdamerow, line 102: see my comment below on keeping all of the objects in memory.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/migration/impl/ModelMapperUtil.java
jdamerow: this is try for all the other methods as well.

jdamerow (changes requested):
All the managers need to use the new repositories or the migration won't be of much use.

jdamerow:
Basically before you run the migration there shouldn't be anything to see, afterwards you should see all concepts again.

PR #209, opened 2020-09-08
Bug/ccp 226

kkrish39:


nowke (approved):

jdamerow:
Make it so, Jenkins.

jdamerow:
a couple things:
- "Search on description" should be "Search in description field" and there should be no period afterwards
- when "Search in description field" is checked, word **and** description fields should be searched, not just description
- the pagination doesn't seem to pass the search in description flag

jdamerow:
Resolve conflicts please

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/core/impl/ConceptManager.java
```diff
@@ -146,14 +146,17 @@ public ConceptEntry getWordnetConceptEntry(String wordnetId) throws LuceneExcept
...
                     throws LuceneException, IllegalAccessException, IndexerRunningException {
         if (pos == null)
             return null;
 
-        Map<String, String> fieldMap = new HashMap<String, String>();
+        Map<String, String> fieldMap = new HashMap<>();
+        
         fieldMap.put(SearchFieldNames.WORD, word);
+        if(isSearchOnDescription)
+            fieldMap.put(SearchFieldNames.DESCRIPTION, word );
```
jdamerow, line 159: please follow our code style

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
something is wrong now (I think it's the use of the TermQuery). If I search for Einstein and the description checkbox is checked, it only finds one result (probably the one that has einstein in the description) and none of the others.

kkrish39:
Can you check once? I searched for einstein with the checkbox checked. I got three rows.
Screen Shot 2020-09-16 at 2 17 15 PM

jdamerow:
Make it so, Jenkins.

jdamerow:
It's still only returned one result when searching on description. Did you reindex after you did the changes to the query?

kkrish39:
> It's still only returned one result when searching on description. Did you reindex after you did the changes to the query?

Let me check once again!

kkrish39:
> It's still only returned one result when searching on description. Did you reindex after you did the changes to the query?

I reindexed and tried once again. Without checkbox checked.
Screen Shot 2020-10-27 at 11 58 27 AM

With checkbox checked.
Screen Shot 2020-10-27 at 11 58 41 AM

jdamerow:
Exactly, the second search should not return fewer results but more. Why does it not find "Einstein's special theory of relativity"?

kkrish39:
> Exactly, the second search should not return fewer results but more. Why does it not find "Einstein's special theory of relativity"?

I see what's happening and what is required. When you said the query should search "concept and description" I thought the combined search of both. 

Now the query is returning only the rows which contain the search term in both concept and description.

jdamerow:
ah, yes, it should be "OR" not "AND"

kkrish39:
I'm pretty sure it can be handled a bit more graciously. Let me know your comments.

ashmichheda (approved):

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java
```diff
@@ -533,7 +535,7 @@ public void indexConcepts(String userName)
                 String searchString = fieldMap.get(search.fieldName());
                 if (searchString != null) {
                     searchString = searchString.toLowerCase();
-                    buildQuery(occur, perFieldAnalyzerWrapper, qBuild, builder, luceneFieldAnnotation, searchString);
+                    buildQuery(occur, perFieldAnalyzerWrapper, qBuild, builder, luceneFieldAnnotation, searchString, isSearchOnDescription);
```
jdamerow, line 538: isn't the only thing you need to do, test if `isSearchOnDescription` is true and if search is the description field, then exclude or not exclude it here?

jdamerow: something like 
```
if (search.fieldName != SearchFieldNames.DESCRIPTION || isSearchOnDescription ) { // or however you get the filed name
 buildQuery(...)
}
```

jdamerow (changes requested):

PR #215, opened 2021-01-04
Bug/ccp 232

kkrish39:
Need to fix Integration tests. Please let me know If the working is as expected.

### DIFF
--- a/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java
+++ b/Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java
@@ -532,6 +532,9 @@ public class LuceneUtility implements ILuceneUtility {
             SearchField search = field.getAnnotation(SearchField.class);
             LuceneField luceneFieldAnnotation = field.getAnnotation(LuceneField.class);
             if (search != null) {
+                if(isSearchOnDescription && search.fieldName().equals("description")){
+                    continue;
+                }
                 String searchString = fieldMap.get(search.fieldName());
                 if (searchString != null) {
                     searchString = searchString.toLowerCase();
@@ -596,19 +599,19 @@ public class LuceneUtility implements ILuceneUtility {
             String searchString, boolean isSearchOnDescription) {
         if (luceneFieldAnnotation.isTokenized()) {
             BooleanQuery.Builder tokenizedQueryBuilder = new BooleanQuery.Builder();
-            buildTokenizedOrWildCardQuery(luceneFieldAnnotation, searchString, qBuild, tokenizedQueryBuilder);
+            buildTokenizedOrWildCardQuery(luceneFieldAnnotation, searchString, qBuild, tokenizedQueryBuilder, isSearchOnDescription);
 
             if (luceneFieldAnnotation.isShortPhraseSearchable()) {
                 BooleanQuery.Builder rootQueryBuilder = new BooleanQuery.Builder();
                 rootQueryBuilder.add(tokenizedQueryBuilder.build(), Occur.SHOULD);
                 // Short word searching
                 BooleanQuery.Builder shortWordSearchQueryBuilder = new BooleanQuery.Builder();
-                shortWordSearchQueryBuilder.add(new PhraseQuery(luceneFieldAnnotation.lucenefieldName() + LuceneFieldNames.UNTOKENIZED_SUFFIX, searchString), Occur.SHOULD);
+                shortWordSearchQueryBuilder.add(new PhraseQuery(luceneFieldAnnotation.lucenefieldName() + LuceneFieldNames.UNTOKENIZED_SUFFIX, searchString), Occur.MUST);
                 
                 /*If the searchDescription is enabled, add the query along with Word Query.
                  * Rightnow isShortPhraseSearchable is enabled only for Word field which avoids multiple entry of search params */
                 if(isSearchOnDescription) {
-                    shortWordSearchQueryBuilder.add(new PhraseQuery(SearchFieldNames.DESCRIPTION, searchString), Occur.SHOULD);
+                    shortWordSearchQueryBuilder.add(new PhraseQuery(SearchFieldNames.DESCRIPTION, searchString), Occur.MUST);
                 }
                 
                 rootQueryBuilder.add(shortWordSearchQueryBuilder.build(), Occur.SHOULD);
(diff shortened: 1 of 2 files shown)


Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java
```diff
@@ -532,6 +532,9 @@ public void indexConcepts(String userName)
             SearchField search = field.getAnnotation(SearchField.class);
             LuceneField luceneFieldAnnotation = field.getAnnotation(LuceneField.class);
             if (search != null) {
+                if(isSearchOnDescription && search.fieldName().equals("description")){
```
jdamerow, line 535: what is this if for? Shouldn't it be if `!isSearchOnDescription`?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java
```diff
@@ -596,19 +599,19 @@ private void buildQuery(BooleanClause.Occur occur, PerFieldAnalyzerWrapper perFi
...
                 // Short word searching
                 BooleanQuery.Builder shortWordSearchQueryBuilder = new BooleanQuery.Builder();
-                shortWordSearchQueryBuilder.add(new PhraseQuery(luceneFieldAnnotation.lucenefieldName() + LuceneFieldNames.UNTOKENIZED_SUFFIX, searchString), Occur.SHOULD);
+                shortWordSearchQueryBuilder.add(new PhraseQuery(luceneFieldAnnotation.lucenefieldName() + LuceneFieldNames.UNTOKENIZED_SUFFIX, searchString), Occur.MUST);
                 
                 /*If the searchDescription is enabled, add the query along with Word Query.
                  * Rightnow isShortPhraseSearchable is enabled only for Word field which avoids multiple entry of search params */
                 if(isSearchOnDescription) {
-                    shortWordSearchQueryBuilder.add(new PhraseQuery(SearchFieldNames.DESCRIPTION, searchString), Occur.SHOULD);
+                    shortWordSearchQueryBuilder.add(new PhraseQuery(SearchFieldNames.DESCRIPTION, searchString), Occur.MUST);
```
jdamerow: why is this `MUST`? Shouldn't it be 'or', which is `SHOULD` I believe?

kkrish39: I believe SHOULD or MUST will not alter the results as we another query combining the word and description query with 'SHOULD'.
``` rootQueryBuilder.add(shortWordSearchQueryBuilder.build(), Occur.SHOULD); ```

File: Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java (hunk 1 above)
Comment (kkrish39) [reply Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java:535]: We shouldn't add the "description" search field when it's searchOnDescription. But we need to add the rest of the query params.

Comment (jdamerow) [reply Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java:535]: I understand that from the code, but why?

File: Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java (hunk 2 above)
Comment (jdamerow) [reply Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java]: why are you changing it then?

**Comment by jdamerow** [comment]
Actually, I'm getting more confused. This ticket is about "Search seems to not find a term if a very short word is included (it only finds it when it is an exact match)". Why does this involved searchOnDescription?

File: Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java (hunk 1 above)
Comment (kkrish39) [reply Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java:535]: A redundant query will be added to the description field right? 

Like, once it will be added through this loop. Another query will be added when the "word" field is trying to build a query.

**Comment by kkrish39** [comment]
Ahh, my bad. I think I mixed up between CCP-226 and CCP-232. I've pushed to changes related to this ticket to CCP-226.

File: Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java (hunk 1 above)
Comment (jdamerow) [reply Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java:535]: See comment below

File: Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java [hunk 3]
```diff
@@ -625,12 +628,17 @@ private void buildQuery(BooleanClause.Occur occur, PerFieldAnalyzerWrapper perFi
...
             if (luceneFieldAnnotation.isWildCardSearchEnabled()) {
                 BooleanQuery.Builder analyzedBuilder = new BooleanQuery.Builder();             
                 createWildCardSearchQuery(luceneFieldAnnotation, searchString, analyzedBuilder, Occur.SHOULD);
                 analyzedBuilder.add(new BooleanClause(
-                        (qBuild.createPhraseQuery(luceneFieldAnnotation.lucenefieldName(), searchString)), Occur.SHOULD));
+                        (qBuild.createBooleanQuery(luceneFieldAnnotation.lucenefieldName(), searchString)), Occur.SHOULD));
+                if(isSearchOnDescription) {
+                    analyzedBuilder.add(new BooleanClause(
+                        (qBuild.createBooleanQuery(SearchFieldNames.DESCRIPTION, searchString)), Occur.SHOULD));
+                }
```
jdamerow, line 640: why is this added here? this would be added multiple times (for every field that is wild card search enabled). There is a similar block in `buildQuery` that is not in develop but doesn't show up as changed. What's up with that?

jdamerow (changes requested):

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/lucene/impl/LuceneUtility.java
kkrish39: The reason is that because the word and description query need to be combined as one.  We are adding Word and description queries to the `analyzedBuilder`  BooleanQuery.Builder object. And you are right, it will be added for every wild card search, but it will work in our case because only the word is enabled for a wildcard search.


Ah, I'm merging the current branch to the CCP-226, not to the develop. As I told you in the above comment, I messed up while pushing commits. It got mixed between CCP-226 and CCP-232. That's the reason why it's not showing up here.

jdamerow: I think we need to make some changes to the search. This doesn't make sense and would create odd queries when used e.g. by the rest endpoint search. There are fields that make sense to search and fields that are limiting the search. I would add another search method (leave the current one the way it was (except refactoring) so that the rest search still works). The new search method takes POS and concept list as arguments (each could be more than just one value). If those are set than the query should have a `AND conceptlist in [...]`/`AND POS in [...]`. All fields that are actually search should be combined with the given operator (in case of the front end search `OR`). Does this make sense? It should be easy then also to add a search to the concept lists.

PR #216, opened 2021-01-05
[CCP-241] Fixed: Non-ASCII characters are not properly getting displayed

kkrish39:
We also need to make sure that the database is also in UTF-8 encoding.


jdamerow:
Make it so, Jenkins.

jdamerow:
Test cases are failing.  I assume it's the other branch, so let's reopen the PR when those are fixed.

PR #218, opened 2021-01-27
Feature/ccp-248-2

kkrish39:
Creating a new PR to merge CCP-248-2 to develop


Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/db/TypeDatabaseClient.java
```diff
@@ -1,153 +1,87 @@
...
-        int endIndex = startIndex + pageSize;
-        if (endIndex > conceptTypes.size()) {
-            endIndex = conceptTypes.size();
-        }
+    public List<ConceptType> getAllTypes(int page, int pageSize, final String sortBy, int sortDirection){
+        page = page < 0 ? 0 : page;
 
-        return new ArrayList<ConceptType>(conceptTypes.subList(startIndex, endIndex));
-    }
+        pageSize = pageSize == 0 ? defaultPageSize : pageSize;
```
jdamerow: should be `< 0`

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/manager/impl/AlternativeIdService.java
```diff
@@ -36,24 +36,19 @@ public void addAlternativeIds(String queriedId, ConceptEntry entry) {
         }
 
         if (conceptTypesService.getConceptTypeByConceptId(queriedId) == IdType.GENERIC_WORDNET_CONCEPT_ID) {
-            entry.getAlternativeIds().add(queriedId);
+            entry.setAlternativeIds(queriedId);
```
jdamerow: still same as before, why are you setting the id here instead of adding it?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/manager/impl/AlternativeIdService.java
```diff
@@ -36,24 +36,19 @@ public void addAlternativeIds(String queriedId, ConceptEntry entry) {
...
         }
         // Specific Wordnet id is added irrespective of what is queried for
         if (entry.getWordnetId() != null) {
-            String[] wordNetIds = entry.getWordnetId().split(",");
-            for (String wordNetId : wordNetIds) {
-                entry.getAlternativeIds().add(wordNetId.trim());
-            }
+            entry.setAlternativeIds(entry.getWordnetId());
```
jdamerow: see above

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/model/ConceptEntry.java
```diff
@@ -0,0 +1,301 @@
...
+
+import edu.asu.conceptpower.app.constants.LuceneFieldNames;
+import edu.asu.conceptpower.app.constants.SearchFieldNames;
+import edu.asu.conceptpower.app.reflect.LuceneField;
+import edu.asu.conceptpower.app.reflect.SearchField;
+
+/**
+ * This model helps to keep track of information about a specific concept
+ * 
+ * @author Keerthivasan Krishnamurthy
```
jdamerow: this is still the same class as before, right? just a different package? Let's put at least Digital Innovation Group  as author as well if there was no author information there before.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/model/ConceptList.java
```diff
@@ -0,0 +1,56 @@
...
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+
+/**
+ * This model helps to keep track of information about a specific concept list
+ * 
+ * @author Keerthivasan Krishnamurthy
```
jdamerow: same as before

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/model/ConceptType.java
```diff
@@ -0,0 +1,103 @@
...
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+/**
+ * 
+ * This model helps to keep track of information about a specific concept type
+ * 
+ * @author Keerthivasan Krishnamurthy
```
jdamerow: see above

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/model/Token.java
```diff
@@ -0,0 +1,80 @@
...
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.JoinColumn;
+import javax.persistence.OneToOne;
+
+/**
+ * This model stores information about the generated tokens for a user
+ * 
+ * @author Keerthivasan Krishnamurthy
```
jdamerow: see above

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/model/User.java
```diff
@@ -0,0 +1,105 @@
...
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+/**
+ * This model helps to keep track of information about the list of users for authentication/authorization
+ * 
+ * @author Keerthivasan Krishnamurthy
```
jdamerow: see above

jdamerow (changes requested):

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/manager/impl/AlternativeIdService.java
kkrish39: It's because the alternativeIds param is a string and we are checking for duplicate while adding an id and appending it to the string. I think I need to rename the setter method.

jdamerow: yes, I believe that was one of my first comments on this ;)

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/model/ConceptEntry.java
```diff
@@ -0,0 +1,308 @@
...
+        if(this.alternativeIds != null) {
+            String[] ids = this.alternativeIds.split(",");
+            return new HashSet<>(Arrays.asList(ids));
+        }
+        return new HashSet<>();
+    }
+    
+    public void addAlternativeId(String alternativeId) {
+        
+        if(alternativeId == null) return;
```
jdamerow, line 289: code style

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/model/ConceptEntry.java
```diff
@@ -0,0 +1,308 @@
...
+        }else {
+            /*This loop is to handle alternativeId parameter with one or more alternativeIds*/
+            for(String altId : alternativeId.split(",")) {
+                altId = altId.trim();
+                this.alternativeIds = (altId.length() > 0 && checkDuplicate(altId)) ? this.alternativeIds : this.alternativeIds+","+altId;
+            }
+        }
+    }
+    
+    private boolean checkDuplicate(String alternativeId) {
```
jdamerow, line 302: this should be called `isDuplicate` so that's its clear what it returns (true if it is a duplicate otherwise false). It's a lot harder to read and understand the above code if you have to look up what the method returns.

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/model/ConceptEntry.java
```diff
@@ -0,0 +1,308 @@
...
+    
+    @Column(name="modified_user")
+    private String modifiedUser;
+
+    @OneToMany( cascade = CascadeType.ALL, fetch= FetchType.EAGER)
+    @JoinColumn(name = "concept_entry_id", referencedColumnName="concept_id")
+    private List<ChangeEvent> changeEvents = new ArrayList<>();
+     
+    @Column(name="alternative_ids")
+    private String alternativeIds;
```
jdamerow, line 122: why did you change this to a string? A lot of the code won't work anymore (see comment below).

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/model/ConceptEntry.java
```diff
@@ -0,0 +1,308 @@
...
+    public void setMergedIds(String mergedIds) {
+        this.mergedIds = mergedIds;
+    }
+    
+    public Set<String> getAlternativeIds() {
+        if(this.alternativeIds != null) {
+            String[] ids = this.alternativeIds.split(",");
+            return new HashSet<>(Arrays.asList(ids));
+        }
+        return new HashSet<>();
```
jdamerow, line 284: there are several places in the code that will call this method to add new ids. This won't work any longer, new ids will be added to the hashset but the hashset is not persisted. Why did you change that?

Conceptpower+Spring/src/main/java/edu/asu/conceptpower/app/service/impl/ConceptMergeService.java
```diff
@@ -201,7 +201,7 @@ private void fillConceptEntry(ConceptEntry entry, ConceptsMergeBean conceptMerge
         if (entry.getAlternativeIds() != null) {
             entry.getAlternativeIds().addAll(conceptMergeBean.getAlternativeIds());
```
jdamerow, line 202: see comments above, this won't persist the changes any longer

jdamerow (changes requested):

